PR #24, opened 2022-01-11
Init add equinix metal

DavidSpek:


michaeljguarino (commented):
Is there a better image for this?  I suspect this will work poorly as an icon.  (Maybe we should just use the equinix icon itself?)

DavidSpek:
@michaeljguarino I've changed the icon

michaeljguarino (approved):

PR #54, opened 2022-06-08
Handle console migration wait flakiness

michaeljguarino:
## Summary

There are a few issues with the way the console is currently deployed.

* we explicitly wait for migration jobs to complete, which is fine on first deploy, but can become an issue if a console pod is killed later and the job is reaped beforehand
* we use statefulsets which were necessary when attempting to implement leader election, but a constant headache when these restarts hit

This replaces the migration wait with a pg wait and replaces the statefulset with a deployment

## Test Plan
ran against our cluster using `plural link`

## Checklist

### DIFF
--- a/rel/config/console.exs
+++ b/rel/config/console.exs
@@ -22,7 +22,7 @@ config :libcluster,
       config: [
         mode: :ip,
         kubernetes_node_basename: "console",
-        kubernetes_selector: "app.kubernetes.io/name=console",
+        kubernetes_selector: "app=console",
         kubernetes_namespace: get_env("NAMESPACE"),
         polling_interval: 10_000
       ]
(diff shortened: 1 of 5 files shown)


rel/config/console.exs
```diff
@@ -22,7 +22,7 @@ config :libcluster,
       config: [
         mode: :ip,
         kubernetes_node_basename: "console",
-        kubernetes_selector: "app.kubernetes.io/name=console",
```
DavidSpek, line 25: I think this is actually the preferred notation over `app=console`. Could also make this change in the helm chart if needed otherwise.

michaeljguarino: they're basically the same, I think.  The service has both labels set

PR #150, opened 2022-11-18
Add better templating for repo encryption key

michaeljguarino:
## Summary

If you've shared a repo, we don't template in the enc key anymore, which can be a weird state if someone's shared prior to a deploy

## Test Plan
local link

## Checklist

### DIFF
--- a/plural/helm/console/values.yaml.tpl
+++ b/plural/helm/console/values.yaml.tpl
@@ -68,9 +68,8 @@ secrets:
   identity: {{ readFile $identity | quote }}
 {{ else if ne (dig "console" "secrets" "identity" "default" .) "default" }}
   identity: {{ .console.secrets.identity | quote }}
-{{ else }}
-  key: {{ readFile (homeDir ".plural" "key") | quote }}
 {{ end }}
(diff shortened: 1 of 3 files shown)


plural/helm/console/values.yaml.tpl
```diff
@@ -68,9 +68,8 @@ secrets:
   identity: {{ readFile $identity | quote }}
 {{ else if ne (dig "console" "secrets" "identity" "default" .) "default" }}
   identity: {{ .console.secrets.identity | quote }}
-{{ else }}
-  key: {{ readFile (homeDir ".plural" "key") | quote }}
 {{ end }}
+  key: {{ dedupe . "console.secrets.key" (readFile (homeDir ".plural" "key")) | quote }}
```
DavidSpek, line 72: Should this be deduped, or always read from the file? The previous logic would always read it from the file.

DavidSpek: Also, what if the `.plural/key` file doesn't exist. You'd need to do a check if that file exists first before doing this line of templating.

michaeljguarino: there are cases where it is not actually in the file.  Also i don't think you can really run `plural build` w/ the key existing, but the function will just return an empty string if it doesn't

PR #182, opened 2023-01-06
Better oauth error handling

michaeljguarino:
## Summary
This would choke opaquely before, a proper error alert is a far better ux (we should change with redesign obviously)

## Test Plan
tested locally

## Checklist

### DIFF
--- a/assets/src/components/OauthCallback.js
+++ b/assets/src/components/OauthCallback.js
@@ -1,7 +1,7 @@
 import { useEffect } from 'react'
 import gql from 'graphql-tag'
 import { Box } from 'grommet'
-import { GqlError } from 'forge-core'
+import { Alert, AlertStatus, GqlError } from 'forge-core'
 import { useMutation } from 'react-apollo'
 import { useLocation } from 'react-router'
 import qs from 'query-string'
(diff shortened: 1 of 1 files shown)


dherault (approved):
More forge-core/grommet should be avoided but I still validate the code.

assets/src/components/OauthCallback.js
```diff
@@ -11,16 +11,36 @@ import { setToken } from '../helpers/auth'
...
+            status={AlertStatus.ERROR}
+            header={error}
+            description={description || "You cannot log into this console instance, make sure your plural user was added to its oidc provider"}
+          />
+        </Box>
+      </Box>
+    </LoginPortal>
+  )
+}
+
```
dogmar, line 40: Wanna use some `<Div>` and/or `<Flex>` from honorable since we're trying to migrate off grommet?

assets/src/components/OauthCallback.js
```diff
@@ -11,16 +11,36 @@ import { setToken } from '../helpers/auth'
...
+    <LoginPortal>
+      <Box gap="medium">
+        <Box
+          gap="xsmall"
+          align="center"
+        >
+          <Alert
+            status={AlertStatus.ERROR}
+            header={error}
+            description={description || "You cannot log into this console instance, make sure your plural user was added to its oidc provider"}
```
dherault, line 33: Lint + Plural with a capital P, OIDC can be capitalized too.

assets/src/components/OauthCallback.js
michaeljguarino: I think we need to rewrite this in the redesign anyway so will probably just punt for now

PR #186, opened 2023-01-10
Container Shell/Terminal

dogmar:


assets/src/components/terminal/TerminalInformation.tsx
```diff
@@ -0,0 +1,111 @@
+import { useState } from 'react'
```
michaeljguarino, line 1: Why is this file here?

maciaszczykm (approved):
@dogmar LGTM, feel free to merge. I'm not merging it now as there should be no conflicts with my profile changes that I'll be doing today and there is one comment from Michael.

assets/src/components/terminal/TerminalInformation.tsx
dogmar: Shouldn't be. It's gone.

PR #251, opened 2023-02-06
fix: Make sure key:value pairs display values on Components metadata page

dogmar:
---
  - https://app.asana.com/0/0/1203894855330888

### DIFF
--- a/assets/src/components/cluster/LabelsAnnotations.tsx
+++ b/assets/src/components/cluster/LabelsAnnotations.tsx
@@ -1,7 +1,7 @@
 import { Card, CardProps, Chip } from '@pluralsh/design-system'
 import { Div, Flex } from 'honorable'
 import { ReactNode } from 'react'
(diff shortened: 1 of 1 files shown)


michaeljguarino (approved):

assets/src/components/cluster/LabelsAnnotations.tsx
```diff
@@ -32,8 +32,13 @@ export function LabelsAnnotationsRow({
   )
 }
 
-export function LabelsAnnotationsTag({ name }) {
-  return <Chip>{name}</Chip>
+export function LabelsAnnotationsTag({ name, value }: LabelPair) {
+  return (
+    <Chip>
+      {name}
+      {value && `: ${value}`}
```
michaeljguarino, line 39: these will both end up bolded right?

dogmar: yup, both bold.

PR #266, opened 2023-02-08
feat: Truncate and reduce label sizes + Responsive enhancements

dogmar:
---
  - https://app.asana.com/0/0/1203837229973050
  - https://app.asana.com/0/1203642829646384/1203896774152795/f


floreks (approved):
Just some linting issue that needs to be fixed, otherwise LGTM

assets/src/components/cluster/LabelsAnnotations.tsx
```diff
@@ -34,45 +36,58 @@ export function LabelsAnnotationsRow({
...
-} & CardProps) {
+  } & CardProps) {
+  const theme = useTheme()
+
   return (
-    <Card
-      padding="xlarge"
-      {...props}
-    >
+    <MetadataCard {...props}>
```
maciaszczykm, line 64: The card will be empty if there are no labels and annotations?

dogmar: Yeah, I guess I should hide the card completely in that case.

dogmar: Doesn't render card if there's no data at all.

PR #277, opened 2023-02-10
fix target name

zreigz:
## Summary
Fix target name for the Proxy object

## Checklist

### DIFF
--- a/plural/helm/console/templates/proxies.yaml
+++ b/plural/helm/console/templates/proxies.yaml
@@ -24,7 +24,7 @@ metadata:
 {{ include "console.labels" . | indent 4 }}
 spec:
   type: sh
-  target: statefulset/console
+  target: statefulset/plural-console
   shConfig:
     command: /opt/app/bin/console
     args: [remote_console]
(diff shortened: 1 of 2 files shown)


plural/helm/console/templates/proxies.yaml
```diff
@@ -24,7 +24,7 @@ metadata:
 {{ include "console.labels" . | indent 4 }}
 spec:
   type: sh
-  target: statefulset/console
```
DavidSpek, line 27: This should actually be `deployment/console`. The console used to be a statefulset but it was changed to a deployment. The `plural-console` statefulset is the postgres database.

DavidSpek (changes requested):
The current proxy is indeed broken but it needs a different fix. Also, can you bump the helm chart version as well.

plural/helm/console/templates/proxies.yaml
zreigz: done

DavidSpek (approved):
LGTM

PR #307, opened 2023-03-01
fix: Editing Groups was crashing

dogmar:
This also brings the editing of groups in-line with the functionality on app.plural.sh

### DIFF
--- a/assets/src/components/account/groups/Group.tsx
+++ b/assets/src/components/account/groups/Group.tsx
@@ -1,13 +1,20 @@
-import { useMutation } from '@apollo/client'
-import { Box } from 'grommet'
-import { ListBoxItem } from '@pluralsh/design-system'
+import {
+  GearTrainIcon,
+  GlobeIcon,
+  IconFrame,
+  PeopleIcon,
+} from '@pluralsh/design-system'
 import { useContext, useState } from 'react'
 
 import { Confirm } from 'components/utils/Confirm'
 
-import { MoreMenu } from 'components/utils/MoreMenu'
-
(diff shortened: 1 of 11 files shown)


assets/src/components/account/groups/GroupEdit.tsx
```diff
@@ -1,195 +1,227 @@
...
+  }, [])
+
+  useEffect(() => {
+    if (open && open !== prevOpen) {
+      reset()
+    }
+  })
+
+  useEffect(() => {
+    console.log('MESSAGe', error?.graphQLErrors[0].message)
```
floreks, line 73: Leftover?

assets/src/components/account/groups/Group.tsx
```diff
@@ -1,103 +1,127 @@
...
-          />
-        ))}
-      </MoreMenu>
+      <Flex
+        flex={false}
+        direction="row"
+        gap="large"
+        align="center"
+      >
+        {group.global && <GlobeIcon size={20} />}
```
maciaszczykm, line 61: I think this was not accessible in console before. Only in the app.plural.sh.

maciaszczykm: cc @dogmar

dogmar: Oh good catch. I thought I had removed all references to global, but I guess i missed one.

maciaszczykm: It will not show anywhere, so not an issue anyways.

PR #309, opened 2023-03-03
Console helm fixes

michaeljguarino:
## Summary
The way we were reading config from the plural dir is incompatible with service account auto-assume.  This will fix for now. Also have console read repo url from context.yaml when present (helps with migrating repos)

## Test Plan
local link

## Checklist


DavidSpek (approved):
Didn’t test myself but approving to unblock.

plural/helm/console/templates/_helpers.tpl
```diff
@@ -48,6 +48,14 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 console-migration-{{ .Release.Revision }}
 {{- end -}}
 
+{{- define "plural-config" -}}
```
michaeljguarino, line 51: shouldn't this name be something like `console.plural-config`?

PR #320, opened 2023-03-08
feat: Convert Groups queries to generated TS queries

dogmar:
Removing an unnecessary ‘global’ icon from the groups list led to updating all the Groups queries.


michaeljguarino (approved):

assets/src/components/account/groups/GroupsList.tsx
```diff
@@ -28,7 +27,7 @@ export function GroupsList({ q }: any) {
       flexGrow={1}
       maxHeight="max-content"
     >
-      {edges?.length > 0 ? (
+      {edges && edges?.length > 0 ? (
```
maciaszczykm, line 30: @dogmar Is this change required? Perhaps we can use `lodash/isEmpty`?

assets/src/components/utils/BindingInput.tsx
```diff
@@ -31,14 +31,26 @@ const FETCHER = {
...
+export function fetchGroups(client: ApolloClient<any>, query, setSuggestions) {
+  client
+    .query<SearchGroupsQuery, SearchGroupsQueryVariables>({
+      query: SearchGroupsDocument,
+      variables: { q: query },
+    })
+    .then(({ data }) => data?.groups?.edges?.map(edge => ({
+      value: edge?.node,
+      label: groupSuggestion(edge?.node),
+    })))
```
maciaszczykm, line 53: @dogmar On a side note, I like your code formatting but it looks like it's your own tool. Perhaps you could add it to our setup somehow if others agree? I mean it's a lot easier to read above code after your changes.

cc @pluralsh/frontend

dogmar: I've been using [prettier](https://prettier.io/) and I'd love to integrate it into our setup if everyone else likes it. It's very opinionated, but I love just not having to think about formatting at all and letting it do its thing.

maciaszczykm: With @floreks we were using prettier in our previous projects. Mostly not directly but with [gts](https://github.com/google/gts). It can serve as a base config but we might prepare something on our end.

And yes, I'd love to have consistent formatting no matter who is doing the changes.

PR #326, opened 2023-03-10
Proposal: Use `prettier` to format front-end code

dogmar:
## Checklist


assets/package.json
```diff
@@ -11,15 +11,18 @@
...
+    "fix": "yarn prettier-write && yarn fix:js && yarn fix:css",
     "fix:js": "eslint ./src --fix",
     "fix:css": "stylelint src/**/*.css --fix",
     "e2e:start": "concurrently -kill-others --success first \"yarn start\" \"wait-on -c waitOnConfig.json -v https-get://localhost:3000 && cd e2e/ && npx cypress run\"",
     "e2e:open": "concurrently -kill-others --success first \"yarn start\" \"wait-on -c waitOnConfig.json -v https-get://localhost:3000 && cd e2e/ && npx cypress open\"",
-    "graphql:codegen": "graphql-codegen --config codegen.yml"
+    "graphql:codegen": "graphql-codegen --config codegen.yml",
+    "prettier-write": "yarn prettier --write ./src",
+    "prettier-check": "yarn prettier --check ./src",
+    "postinstall": "cd .. && husky install assets/.husky"
```
maciaszczykm, line 25: We can run it automatically like [here](https://github.com/kubernetes/dashboard/blob/master/modules/web/.yarnrc.yml#L9) with [postinstall](https://github.com/kubernetes/dashboard/blob/master/modules/web/.yarn/plugins/%40yarnpkg/plugin-postinstall.cjs) plugin.

assets/src/components/account/groups/Group.tsx
```diff
@@ -30,29 +34,21 @@ export default function Group({ group, q }: { group: GroupT; q: any }) {
...
         text={group.name}
         description={group.description || 'no description'}
       />
-      <Flex
-        flex={false}
-        direction="row"
-        gap="large"
-        align="center"
-      >
+      <Flex flex={false} direction="row" gap="large" align="center">
```
maciaszczykm, line 51: One thing to consider is keeping args in separate lines. I think it was easier to read. What do you think?

floreks: Agree, I also prefer arg-per-line style.

floreks:
Great idea in general. We should use an external formatter to avoid colliding with each other's IDEs.

maciaszczykm:
@dogmar My recent changes in master branch have caused a lot of conflicts in this pull request, so I have resolved them. I have also added configuration files for prettier with `singleAttributePerLine: true`. This is just a proposal, let me know what you think and feel free to revert any changes that I have made here.

assets/package.json
dogmar: Oh great! I was struggling to find a way to get husky installed automatically with yarn. I'll take a look at this plugin.

dogmar:
I missed that `singleAttributePerLine: true` was even an option. I prefer it as well. However, I think we should stick with plural's default for `printWidth`. They have an explanation [here](https://prettier.io/docs/en/options.html#print-width). If these rest of you still feel strongly about increasing it, I could be convinced to go to `100`.

maciaszczykm:
@dogmar We can keep 80 for now, and go back to the topic if there will be a need. It is not a strong opinion from my side, rather just me being used to it. Looking at the lines of code stats for this PR (+23k, -51k) the balance is pretty good anyways.

assets/src/components/account/groups/Group.tsx
dogmar: Also makes for cleaner diffs. Looks like we all agree here!

dogmar:
I went ahead and changed this PR so all the commits until the very end leave `assets/src` alone, then rebased to be sure we had the very latest code from `origin/main` before running `yarn fix`. Also added the postinstall yarn plugin to install husky (thanks, @maciaszczykm !). Should be ready for final review.

assets/package.json
```diff
@@ -11,15 +11,19 @@
...
+    "fix": "run-s fix:format fix:js fix:css",
+    "fix:format": "yarn prettier-write",
     "fix:js": "eslint ./src --fix",
     "fix:css": "stylelint src/**/*.css --fix",
     "e2e:start": "concurrently -kill-others --success first \"yarn start\" \"wait-on -c waitOnConfig.json -v https-get://localhost:3000 && cd e2e/ && npx cypress run\"",
     "e2e:open": "concurrently -kill-others --success first \"yarn start\" \"wait-on -c waitOnConfig.json -v https-get://localhost:3000 && cd e2e/ && npx cypress open\"",
-    "graphql:codegen": "graphql-codegen --config codegen.yml"
+    "graphql:codegen": "graphql-codegen --config codegen.yml",
+    "prettier-write": "yarn prettier --write ./src",
+    "prettier-check": "yarn prettier --check ./src"
```
maciaszczykm, line 26: Can you move it to lint:format above?

maciaszczykm: Perhaps we should also rename lint to check?

assets/package.json
```diff
@@ -139,19 +143,28 @@
...
+    "eslint-config-prettier": "8.7.0",
     "eslint-plugin-import": "2.27.5",
     "eslint-plugin-import-newlines": "1.3.1",
     "eslint-plugin-jsx-a11y": "6.7.1",
     "eslint-plugin-react": "7.32.2",
     "eslint-plugin-react-hooks": "4.6.0",
+    "husky": "8.0.3",
     "jsdom": "21.1.0",
+    "lint-staged": "13.2.0",
+    "npm-run-all": "^4.1.5",
```
maciaszczykm, line 155: I forgot to remove ^.

maciaszczykm (approved):
LGTM, with some tiny nit-picks.

assets/package.json
dogmar: fixed

assets/package.json
dogmar: Removed those. I think leaving everything as `lint` is fine and then we don't have to worry about breaking CI.

PR #347, opened 2023-04-27
feat: Update cluster api recipes and terraform

floreks:
## Checklist


plural/terraform/aws-cluster-api/main.tf
```diff
@@ -0,0 +1,59 @@
+resource "kubernetes_namespace" "console" {
```
michaeljguarino, line 1: why can't we just use the existing console terraform modules?  Aren't they effectively the same?

zreigz (approved):

plural/terraform/aws-cluster-api/main.tf
floreks: we probably could, but we still need `deps.yaml` that targets `aws-bootstrap-cluster-api`. Is it possible to use new deps and link existing files?

floreks:
Closing in favour of https://github.com/pluralsh/console/pull/357

PR #348, opened 2023-04-27
fix: Console Pod Logs broken if not in a plural application

dogmar:
## Summary

Created a new 'Logs' tab for Pods pages that have no associated apps. Pods without an associated apps will have the 'View logs' button direct to this tab while pods with apps will still be linked to the App page's logs tab.

## Checklist

---
  - https://app.asana.com/0/0/1203951917789151

### DIFF
--- a/assets/src/components/cluster/pods/PodInfo.tsx
+++ b/assets/src/components/cluster/pods/PodInfo.tsx
@@ -13,6 +13,8 @@ import { POD_INFO_Q } from '../queries'
 import { SubTitle } from '../nodes/SubTitle'
 import { ContainersList } from '../containers/ContainersList'
 
+import { useNamespaceIsApp } from '../../hooks/useNamespaceIsApp'
+
 import PodMetadata from './PodMetadata'
 import PodConditions from './PodConditions'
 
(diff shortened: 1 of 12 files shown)


assets/src/components/cluster/pods/PodInfo.tsx
```diff
@@ -27,14 +29,30 @@ export const statusesToRecord = (statuses?: Maybe<Maybe<ContainerStatus>[]>) =>
...
-function getLogUrl({ name, namespace }) {
-  return `/apps/${namespace}/logs?${asQuery({ pod: name })}`
+function useGetLogUrl({
+  name,
+  namespace,
+}: {
+  name?: string
+  namespace?: string
+}) {
+  const isApp = useNamespaceIsApp(namespace)
```
michaeljguarino, line 39: could we just pass this as an argument and ignore the potentially risky app name fetch?

dogmar: Is there some field on a Pod that I could use to determine if it's an app or not? Otherwise I'd just be pushing the app name fetch further up the render tree.

michaeljguarino: Not easily, I thought you might be able to determine it from context of where the component is called, eg if it's called from an `/apps/` prefixed url use something like that link, otherwise non-app link?

dogmar: Nah, I need to know whether it's an app or not on the Pod page, where it's not obvious.

dogmar: Either that, or we always show the logs on the Pod page, even if there is an associated App page where you could also view the logs.

michaeljguarino (approved):
Let's just try this for now, it should be accurate enough

PR #355, opened 2023-05-17
fix: Add domain uniqueness check to the installer and update plural context query API

floreks:
## Summary

Similar to https://github.com/pluralsh/plural/pull/1102

I have also noticed that `pluralContext` query does not return buckets and domains. I have updated API to support that.

## Test Plan

Locally + updated API unit test

## Checklist

---
  - https://app.asana.com/0/0/1204610425075657

### DIFF
--- a/assets/src/components/contexts/PluralContext.tsx
+++ b/assets/src/components/contexts/PluralContext.tsx
@@ -22,7 +22,7 @@ function PluralProvider({ children }): ReactElement {
     [data]
   )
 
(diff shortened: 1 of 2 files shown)


maciaszczykm (approved):

lib/console/plural/context.ex
floreks, line 1: @michaeljguarino could you take a look at the elixir part?

michaeljguarino (approved):

PR #437, opened 2023-10-10
wip: feat(cd): add create cluster dialog

floreks:
## Summary

```
TBD
```

## Test Plan

Locally

## Checklist


lib/console/schema/cluster.ex
```diff
@@ -51,6 +51,22 @@ defmodule Console.Schema.Cluster do
     end
   end
 
+  defmodule ClusterConfiguration do
```
michaeljguarino, line 54: So I actually think we should put these in CloudSettings since it would allow them to be strictly typed for each cloud.  Perhaps we could also have this for stray stuff but biasing towards typing will keep the system generally more consistent.

You'll then need to write a bit of translation code to flush those settings into the service config

lib/console/deployments/clusters.ex
```diff
@@ -487,13 +487,18 @@ defmodule Console.Deployments.Clusters do
         %{name: "version", value: cluster.version},
         %{name: "nodePools", value: Jason.encode!(node_pools)}
         | credential_config(cluster)
+        | config_attributes(cluster)
```
michaeljguarino, line 490: I think you might need to make this:

` | credential_config(cluster) ++ config_attributes(cluster) ]`

michaeljguarino (approved):
lgtm

assets/src/components/cd/clusters/create/helpers.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+const ProviderToLogo: { readonly [key in Provider]: ComponentType } = {
+  [Provider.AWS]: AwsLogoIcon,
+  [Provider.GCP]: GoogleLogoIcon,
+  [Provider.Azure]: AzureLogoIcon,
+}
+
+const RegionsForProvider: { readonly [key in Provider]: Array<string> } = {
+  [Provider.AWS]: [
```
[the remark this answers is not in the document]
floreks: Ye, that would be great. We have the same hardcoded regions in the plural app.

PR #451, opened 2023-10-17
feat: Component Details path change + Service deployment creation

dogmar:
- Enable creating Service Deployments in the UI
- Change Component Details urls to use IDs
- Show service name in breadcrumbs instead of ID after loading

### DIFF
--- a/assets/src/components/component/ComponentDetails.tsx
+++ b/assets/src/components/component/ComponentDetails.tsx
@@ -52,6 +52,7 @@ export function ComponentDetails({
   pathMatchString: string
   serviceId?: string
 }) {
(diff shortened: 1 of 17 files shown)


assets/src/components/component/ComponentDetails.tsx
```diff
@@ -52,6 +52,7 @@ export function ComponentDetails({
   pathMatchString: string
   serviceId?: string
 }) {
+  console.log('klink')
```
michaeljguarino, line 55: probably should remove some of these console logs

michaeljguarino (approved):

PR #452, opened 2023-10-18
feat: CD frontend improvements

maciaszczykm:
- add delete option for services
- add initial version of clusters metadata tab
- remove duplicated components like pod raw, events etc.
- add missing loading indicator, keep same view on cluster switch and other smaller improvements

---
  - https://app.asana.com/0/0/1205735326658212


assets/src/graph/cdClusters.graphql
```diff
@@ -30,6 +30,14 @@ fragment NodeMetricFragment on NodeMetric {
   window
 }
 
+fragment NodePoolFragment on NodePool {
```
dogmar, line 33: You don't actually need to put "Fragment" at the end of of fragment names. GQL Codegen makes this into a type named `NodePoolFragmentFragment`

maciaszczykm: Ok thanks. I will probably need to refactor this graphql code later to reorganize where fragments are located.

PR #484, opened 2023-11-08
feat: Add link to cluster services

maciaszczykm:
---
  - https://app.asana.com/0/0/1205887360906920


michaeljguarino (approved):

assets/src/components/cd/cluster/ClusterMetadata.tsx
```diff
@@ -80,7 +86,23 @@ function MetadataCard({
...
+              <A
+                as={Link}
+                to={getServiceDetailsPath({
+                  clusterId: cluster?.id,
+                  serviceId: cluster?.service?.id,
+                })}
+                inline
+              >
+                {cluster?.service?.name}
+              </A>
```
dogmar, line 103: Should avoid adding new instances of honorable components. There's `components/utils/typography/InlineLink.tsx`  that could be used.

dogmar (approved):

assets/src/components/cd/cluster/ClusterMetadata.tsx
maciaszczykm: You're right. That was quick copy paste, I will update it.

maciaszczykm: Done, thanks for pointing it out.

PR #487, opened 2023-11-09
feat: Add exec into pod for CD clusters

maciaszczykm:
Zrzut ekranu 2023-11-9 o 13 16 56

---
  - https://app.asana.com/0/0/1205829990058096


michaeljguarino (approved):

assets/src/components/cd/cluster/pod/PodInfo.tsx
```diff
@@ -25,7 +38,34 @@ function ViewLogsButton() {
...
 
+export const ColActions = ({
+  clusterId,
+  podName,
+  namespace,
+}: {
+  clusterId?: string
+  podName?: string
+  namespace?: string
+}) =>
```
dogmar, line 49: I've discovered that creating table columns as functions this way can cause weird effects like Modals closing when the underlying data changes even when memoizing the columns, so I've started passing extra data in through the `meta` prop of react-table. Here's a basic example:

```
export const ColActions = columnHelper.accessor((row) => row.name, {
  id: 'actions',
  cell: function Cell({ table, row }) {
    const { refetch, outsideId } = table.options.meta as any

    return (
      <Button onClick={() => refetch?.()}>
        {outsideId}: {row.original.name}
      </Button>
    )
  },
})

const data = [
  { id: '1', name: 'name1' },
  { id: '1', name: 'name1' },
]

const columns = [ColActions]

export function ATable() {
  return (
    <Table
      data={data}
      columns={columns}
      reactTableOptions={{
        meta: {
          outsideId: 'outside-id',
          refetch: () => {
            console.log('refetch')
          },
        },
      }}
    />
  )
}
```

maciaszczykm: Good to know, I will update it tomorrow.

maciaszczykm: Good to know, thanks. I will update it tomorrow.

PR #499, opened 2023-11-13
feat: Add Getting Started guide to Clusters page

dogmar:
If user has 1 or fewer clusters, show Getting Started Guide below clusters table.


michaeljguarino (commented):
we can PR these against the `master` branch now

michaeljguarino (approved):

PR #523, opened 2023-11-30
feat: Multi-source Helm FE changes

dogmar:
## Checklist

---
  - https://app.asana.com/0/0/1206060022092806


assets/src/components/cd/services/deployModal/DeployServiceSettingsHelmValues.tsx
```diff
@@ -0,0 +1,172 @@
+import { Button, FormField, Input } from '@pluralsh/design-system'
```
michaeljguarino, line 1: So this UI should have two basic things that can be included:

* code editor for a values yaml blob (kind of like what's there currently in there in some other places where we modify helm configuration)
* the ability to specify a list of file names (which would ultimately point to helm values files stored in git)

michaeljguarino (approved):

PR #536, opened 2023-12-05
feat: Display values.yaml file and allow new file on upgrade

dogmar:
## Summary
- Enable editing of helm `values` and `valuesFiles` in three-dots menu and in service details
- Make sure git settings are still editable from three-dots menu when a helm repo is also present
- Enable editing settings (git, helm, delete protection) form service details

## Checklist

---
  - https://app.asana.com/0/0/1206095700447450

### DIFF
--- a/assets/src/components/cd/services/ServiceSettings.tsx
+++ b/assets/src/components/cd/services/ServiceSettings.tsx
@@ -44,7 +44,7 @@ export function ServiceSettings({
   )
 }
 
-function ChartUpdate({ repo, state, updateState }) {
+export function ChartUpdate({ repo, state, updateState }) {
   const { data } = useHelmRepositoryQuery({
     variables: {
       name: repo?.name || '',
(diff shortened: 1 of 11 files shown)


michaeljguarino (commented):
I'm seeing the default code editor for the `helm.values` field having a dummy comment in it, will that cause it to be written to the db or will it be null unless edited?

dogmar:
> I'm seeing the default code editor for the `helm.values` field having a dummy comment in it, will that cause it to be written to the db or will it be null unless edited?

That was a real value from the db I had set during my testing, not a placeholder. I just removed it.

michaeljguarino:
> > I'm seeing the default code editor for the `helm.values` field having a dummy comment in it, will that cause it to be written to the db or will it be null unless edited?
> 
> That was a real value from the db I had set during my testing, not a placeholder. I just removed it.

ok no worries

michaeljguarino (approved):

PR #542, opened 2023-12-07
feat: Add CD sections/clusters to Command Palette

dogmar:
## Checklist


michaeljguarino (commented):
Does it currently let you typeahead to nav to a specific cluster or service or just the overall tab for each of those?  I thought the code might do that but not seeing how to do it in-ui

dogmar:
> Does it currently let you typeahead to nav to a specific cluster or service or just the overall tab for each of those?  I thought the code might do that but not seeing how to do it in-ui

You can type-ahead to specific clusters (you should be able to just type the name of a cluster), but not services, since there could be so many of those. I guess I can look into whether the palette supports async searching.

michaeljguarino:
> > Does it currently let you typeahead to nav to a specific cluster or service or just the overall tab for each of those?  I thought the code might do that but not seeing how to do it in-ui
> 
> You can type-ahead to specific clusters (you should be able to just type the name of a cluster), but not services, since there could be so many of those. I guess I can look into whether the palette supports async searching.

Huh, I don't see the name typeahead working

michaeljguarino (approved):

PR #563, opened 2023-12-21
feat: Lazy-load route components

dogmar:
## Summary

Make most routes load their components lazily to reduce initial bundle size.


michaeljguarino (commented):
how much testing do we have of this?

dogmar:
> how much testing do we have of this?

Not a whole lot. I've started going through the routes and fixing some small things here and there, but we should probably hold off on merging until the new year.

maciaszczykm (approved):

PR #615, opened 2024-01-19
feat: Disable main provider deletion

maciaszczykm:


michaeljguarino (approved):
I actually think there's a `self` field too that signals the base provider as well

PR #629, opened 2024-01-25
feat: update RBAC for secret/configmap and change crd schema

floreks:
## Summary

- Update controllers RBAC configuration to allow Server/ConfigMap patching
- Make repositoryRef optional in ServiceDeployment
- Make providerRef optional in GlobalService
- Add optional `name` field to ServiceDeployment spec. If not provided object name from ObjectMeta will be used.
- Allow providing userEmail/groupName in permissions bindings for cluster/serviceDeployment

Group ID/User ID is backfilled into the CRD when name or email is provided. This is done for backwards compatibility so we can use id or name as needed.
image

## Test Plan

Locally

## Checklist

---
  - https://app.asana.com/0/0/1206407280605439
  - https://app.asana.com/0/0/1206421164635741
  - https://app.asana.com/0/0/1206421164635735


michaeljguarino (commented):
Could we also make a small make script to copy these crds into the two charts appropriately (i think it's still a manual process on release and could easily get forgotten)

michaeljguarino (approved):

michaeljguarino (commented):
I think the fetching of existing services might not currently respect the name-in-spec, but otherwise lgtm

floreks:
> I think the fetching of existing services might not currently respect the name-in-spec, but otherwise lgtm

You're right. I have now replaced all name/namespace usage and used function getter.

michaeljguarino (approved):

.github/workflows/test.yaml
```diff
@@ -1,120 +0,0 @@
-name: CI / Console
-
-on:
-  push:
-    branches: [ master ]
```
michaeljguarino, line 5: should we be removing this file?

floreks: oh wait, good catch. I was testing crd copy make target to make sure it works in GH action and must have accidentally removed this one.

charts/controller/crds/deployments.plural.sh_servicedeployments.yaml
```diff
@@ -73,6 +73,9 @@ spec:
                       type: object
                     type: array
                 type: object
+                x-kubernetes-validations:
```
michaeljguarino, line 76: Bindings should be mutable, I think. It won't happen *that* often but it will occasionally.

floreks: It is possible in general, but we do not have any logic to update bindings on change. It would require running a different `updateRbac` mutation. We would also need to keep separate SHA just for bindings to check if they have changed to update them independently of i.e. Service/Cluster.

I can add an issue to handle RBAC updates and work on that next. Right now any change to bindings would not be reflected in the system. They are only used during initial creation.

michaeljguarino: Ok fair enough.  It's also possible I can add that back to the main mutation, since it's going to be awkward to manage separately everywhere.

floreks: Ye, that would be the best if we could also pass bindings to update mutation for cluster and service. This way it will be pretty much automatically handled by the controller. Only small change to update attributes will be needed.

michaeljguarino: Let's just configure them as mutable and I'll make the console api change in the next release and we should be good.

controller/Makefile
```diff
@@ -107,6 +107,7 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 .PHONY: manifests
 manifests: controller-gen ## generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	@$(MAKE) -s chart-crds
```
floreks, line 110: This should ensure that when we regenerate base crds, controller chart copies are being updated as well.

PR #637, opened 2024-01-28
Backup/Restore data definitions, Service Context CRUD

michaeljguarino:
## Summary
Added basic schema defs for backup/restore, and also implemented a service context api.  The general premise is a reusable, named block of config that can be communicated between tools, eg pulumi/terraform, then bound by name to a service to reconfigure it as needed.

## Test Plan
unit

## Checklist


zreigz:
we also need Get/List methods

PR #640, opened 2024-01-30
feat: add scm connection crd/controller and improve controller reconcile

floreks:
## Summary

- Add `ScmConnection` CRD and controller
- Remove kubebuilder immutable validation from `RepositoryRef` on `ServiceDeployment`
- Add support for `valuesRaw` helm config to `ServiceDeployment`. It is an object that can hold anything. It is transformed into a JSON string before sending to the Console API.
- Extend `Synchronized` condition reason to support `Error` and `NotFound` states. Updated all controllers to use it.
- Add examples for `ServiceDeployment` with raw values config and `ScmConnection`.
- Improve overall controller reconcile loop running multiple times when i.e. owned objects are changed. I have added custom predicates to make sure that CRD is reconciled when generation changes and for related Secrets/ConfigMaps when their resourceVersion changes. It does not impact normal reconciles that happen at regular intervals, only event-based. There were times where creation of Secret/ConfigMap with CRD at the same time could trigger multiple reconciles for the same CRD and result in resource being created in Console and CRD being marked as readonly. This seems to stabilize it, but we should still keep an eye on that.
- Fix an issue where deletion of Console resource synced with readonly CRD instance was causing controller to throw not found errors indefinitely.

## Test Plan

Running locally against https://console.cdaws.onplural.sh/

## Checklist

---
  - https://app.asana.com/0/0/1206469124013240
  - https://app.asana.com/0/0/1206421164635744
  - https://app.asana.com/0/0/1206469124013237


controller/api/v1alpha1/servicedeployment_types.go.go
```diff
@@ -31,6 +32,8 @@ type ServiceHelm struct {
 	// +optional
 	ValuesRef *corev1.ConfigMapKeySelector `json:"values,omitempty"`
 	// +optional
+	ValuesRaw *runtime.RawExtension `json:"valuesRaw,omitempty"`
```
michaeljguarino, line 35: could we just name this `values`?  Would be more what I'd expect as a user.

floreks: Sure, but it will be a breaking change because `values` is already used above for config map ref. Existing CRDs with old spec will have to be updated first.

michaeljguarino: Oh I see why, I wouldn't worry about that breaking change, it's almost not being used explicitly, certainly not by high priority users

PR #649, opened 2024-02-02
feat: automate rbac generation in controller chart and update gh action to use it

floreks:
## Summary

Added `make` target that uses rbac generated by `kustomize` based on `kubebuilder` comments and creates a helm version of it with `helmify`, then copies it to our controller helm chart.

## Test Plan

Locally with `make`

## Checklist


michaeljguarino (approved):

michaeljguarino (commented):
This only will update the rbac resource in the main charts right?  And is the `tmp` folder cleaned up so it won't be committed (or maybe there's a gitignore i missed)?

floreks:
`.gitignore` was updated. `tmp` dir won't be committed. It generates helm to `tmp` dir and only copies `manager-rbac.yaml` file to the controller chart.

PR #656, opened 2024-02-05
feat: add pr automation controller

floreks:
## Summary

- Bump `kubebuilder` to `v.0.14.0`. This cause all CRD schema to slightly change
- Refactor `Status` into a common struct as most of our resources uses the same status object and methods
- Simplify reading our CRD references and their ID
- Add a new `PrAutomation` controller
- Add a sample `pr_automation.yaml`
- Improve controller tests (currently only PrAutomation uses it) to be able to run tests in parallel instead of in order

## Test Plan

Running controller locally on existing `https://console.cd-demo.onplural.sh` API.

## Checklist


michaeljguarino (commented):
I think we need to also add the `creates` field which was fleshed out recently (in addition to `updates`)

controller/api/v1alpha1/prautomation_types.go
```diff
@@ -0,0 +1,318 @@
...
+
+func (in *PrAutomation) Attributes(clusterID *string, serviceID *string, connectionID *string) *console.PrAutomationAttributes {
+	attrs := console.PrAutomationAttributes{
+		Name:          lo.ToPtr(in.ConsoleName()),
+		Identifier:    in.Spec.Identifier,
+		Documentation: in.Spec.Documentation,
+		Title:         in.Spec.Title,
+		Message:       in.Spec.Message,
+		Branch:        in.Spec.Branch,
+		Updates:       in.Spec.Updates.Attributes(),
```
michaeljguarino, line 76: in particular here doesn't pass the creates spec

floreks: oh, I must have missed it

floreks:
> I think we need to also add the `creates` field which was fleshed out recently (in addition to `updates`)

Attributes were also missing repository ID. That should cover it https://github.com/pluralsh/console/pull/656/commits/e8b4a6a0afdab9f34acd24bfc843bd388ba40bfa.

michaeljguarino (approved):

PR #672, opened 2024-02-12
feat: updated compatibilities for istio

smyja:
## Summary
added more compatibilities and their non compatible versions as seen here https://istio.io/v1.12/docs/releases/supported-releases/#support-status-of-istio-releases

## Checklist


static/compatibilities/istio.yaml
```diff
@@ -33,4 +33,28 @@ versions:
 - version: 1.13.0
   kube: ['1.23', '1.22', '1.21', '1.20']
   requirements: []
-  incompatibilities: []
\ No newline at end of file
+  incompatibilities: []
+- version: 1.12
+  kube: ['1.22', '1.21', '1.20', '1.19']
+  requirements: []
+  incompatibilities: ['1.18', '1.17', '1.16']
```
michaeljguarino, line 40: This is useful, but the data structure is actually a bit different, eg each will be a:

```yaml
name: <component-name>
version: <vsn>
```

struct. In this case you could use:

```yaml
name: kubernetes
version: 1.22
```

michaeljguarino: The idea is this would need to be flexible to express incompatibilities between kubernetes operators as well

smyja: > This is useful, but the data structure is actually a bit different, eg each incompatibility entry will be a:
> 
> ```yaml
> name: <component-name>
> version: <vsn>
> ```
> 
> struct. In this case you could use:
> 
> ```yaml
> name: kubernetes
> version: 1.22
> ```

okay, I've updated it. Kindly check to confirm if that's what you mean

michaeljguarino (approved):
lgtm

PR #676, opened 2024-02-12
Cluster Backup/Restore list queries

michaeljguarino:
## Summary
Realized we hadn't implemented these, adding them now

## Test Plan
unit

## Checklist

### DIFF
--- a/lib/console/graphql/deployments/backup.ex
+++ b/lib/console/graphql/deployments/backup.ex
@@ -103,6 +103,8 @@ defmodule Console.GraphQl.Deployments.Backup do
   end
 
   connection node_type: :object_store
+  connection node_type: :cluster_backup
+  connection node_type: :cluster_restore
 
   object :public_backup_queries do
     field :cluster_restore, :cluster_restore do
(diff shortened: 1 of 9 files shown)


lib/console/graphql/deployments/backup.ex
```diff
@@ -142,6 +144,20 @@ defmodule Console.GraphQl.Deployments.Backup do
       resolve &Deployments.resolve_cluster_backup/2
     end
 
+    connection field :cluster_backups, node_type: :cluster_backup do
+      middleware Authenticated
+      arg :cluster_id, non_null(:id)
```
maciaszczykm, line 149: Will there be global list of backups (for all clusters)? I think I saw one in early designs, but I am not sure.

maciaszczykm (approved):

lib/console/graphql/deployments/backup.ex
michaeljguarino: I can make one but is there a need for it?  I'm actually not sure if it makes a lot of sense (could be super noisy)

maciaszczykm: Yep, there could be a lot of backups if it will be configured for multiple clusters. Regarding clusters, how should we find list of clusters with backups enabled? One way is using object stores list and mapping them to clusters, would that be okay?

maciaszczykm: Yep, makes sense.

PR #703, opened 2024-02-21
feat: Add service pod view

maciaszczykm:
Adds service pod view (route: `/cd/clusters/:clusterId/services/:serviceId/pods/:namespace/:name`) that reuses existing components but it is based on pods query with `serviceId` param instead of `clusterId`. This view is referenced only from service component pod lists.

Zrzut ekranu 2024-02-21 o 15 46 20


michaeljguarino (approved):

michaeljguarino:
Think there might be some js linter issues to resolve first though?

PR #747, opened 2024-03-04
feat: Show badge with open PR count in Pipeline Stage nodes

dogmar:
## Summary
Screenshot 2024-03-04 at 11 43 00 AM
Screenshot 2024-03-04 at 11 43 10 AM
Screenshot 2024-03-04 at 11 43 26 AM
Screenshot 2024-03-04 at 11 43 31 AM
Screenshot 2024-03-04 at 11 43 44 AM
Screenshot 2024-03-04 at 11 44 08 AM

### DIFF
--- a/assets/src/components/cd/pipelines/nodes/StageNode.tsx
+++ b/assets/src/components/cd/pipelines/nodes/StageNode.tsx
@@ -153,7 +153,7 @@ function PrsButton({
 }) {
   const [open, setOpen] = useState(false)
   const numOpenPrs = pullRequests.reduce(
-    (acc, pr) => (pr?.status === PrStatus.Open ? acc + 1 : 0),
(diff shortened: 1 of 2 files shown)


michaeljguarino:
Awesome, lgtm

assets/src/components/cd/pipelines/nodes/StageNode.tsx
```diff
@@ -131,25 +132,51 @@ const IconHeadingInnerSC = styled.div(({ theme }) => ({
   width: '100%',
 }))
 
+export const PrCountBadgeSC = styled(CountBadge).attrs(() => ({
```
michaeljguarino, line 135: we should probably have a design system component that can do this at some point.  Grommet had one they called Stack that effectively let you overlay components at different anchor points which is v useful.

PR #754, opened 2024-03-05
Make cluster id immutable for services

michaeljguarino:
## Summary
Not sure if there are actually some other fields worth pinning, there is a justification to allow repository to be changable along w/ other things.

## Test Plan
unit test

## Checklist


maciaszczykm (approved):
Cluster ref seems to be only thing that we have validation for service deployments in controller: https://github.com/pluralsh/console/blob/master/controller/api/v1alpha1/servicedeployment_types.go#L75

We have it for few other fields i.e. cluster or provider cloud.

PR #772, opened 2024-03-12
feat: Implement refresh tokens in UI

dogmar:
## Summary
- Set `refresh-token` cookie on login
- When jwt expiration is within 15min of expiring, call `refresh` with `refresh-token` to renew jwt
- For all gql queries, if it returns with authentication error and `refresh-token` is set, call `refresh` and retry query with new jwt. If `refresh` or retried query errors, clear `refresh-token` and jwt, then logout.

## Checklist


michaeljguarino (approved):
This mostly lgtm but we def need to test this one thoroughly.  I'll also need to follow up w/ a server change to drop jwt lifespan.

PR #777, opened 2024-03-14
feat: Add proper support for ServiceDeployment in components tab

dogmar:
## Summary
Screenshot 2024-03-14 at 2 27 20 PM

## Checklist

### DIFF
--- a/assets/src/components/component/ComponentDetails.tsx
+++ b/assets/src/components/component/ComponentDetails.tsx
@@ -1,6 +1,6 @@
-import { SubTab, TabList, TabPanel } from '@pluralsh/design-system'
+import { Button, SubTab, TabList, TabPanel } from '@pluralsh/design-system'
 import { useContext, useEffect, useMemo, useRef } from 'react'
-import { Outlet, useMatch, useNavigate } from 'react-router-dom'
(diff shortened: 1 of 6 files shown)


michaeljguarino (approved):
sick

assets/src/components/component/ComponentDetails.tsx
```diff
@@ -174,44 +181,58 @@ export function ComponentDetails({
...
+                  <SubTab>{label}</SubTab>
+                </LinkTabWrap>
+              ))}
+            </TabList>
+            {serviceDeploymentServiceId && service?.cluster?.id && (
+              <Button
+                as={Link}
+                to={getServiceDetailsPath({
+                  serviceId: serviceDeploymentServiceId,
+                  clusterId: service?.cluster?.id,
```
michaeljguarino, line 217: This is not guaranteed to be the cluster actually.  If we need both you can actually load the actual `ServiceDeployment` by querying the `reference` object on the service

michaeljguarino: A service crd is usually created on the management cluster, but will be used to create services on any other cluster

PR #806, opened 2024-04-03
Service details sidecar improvements

renemennab:
## Summary
- adds namespace, repository and helmRepository to service details sidecar 
- restricts helm values to admin users

[linear issue](https://linear.app/pluralsh/issue/ENG-1945/service-details-sidecar-improvements)


assets/src/components/cd/services/service/ServiceDetailsSidecar.tsx
```diff
@@ -92,6 +107,11 @@ export function ServiceDetailsSidecar({
             <Chip severity="success">None</Chip>
           )}
         </SidecarItem>
+        {helmRepository && (
+          <SidecarItem heading="Helm Repository">
```
michaeljguarino, line 111: This should ideally include the helm icon like in the service table row

assets/src/components/cd/services/service/ServiceDetailsSidecar.tsx
```diff
@@ -103,6 +123,9 @@ export function ServiceDetailsSidecar({
             {helm.version}
           </SidecarItem>
         )}
+        {repository && (
+          <SidecarItem heading="Git repository">{repository.url}</SidecarItem>
```
michaeljguarino, line 127: same, this should ideally include a git icon

assets/src/components/cd/services/service/ServiceDetailsSidecar.tsx
```diff
@@ -37,7 +37,19 @@ export function ServiceDetailsSidecar({
...
+    status,
+    cluster,
+    git,
+    helm,
+    namespace,
+    repository,
+    helmRepository,
+  } = serviceDeployment
+
+  console.log('repository', repository)
```
michaeljguarino, line 52: rm

assets/src/components/cd/services/service/ServiceDetailsSidecar.tsx
renemennab: I added the icons, I am just not sure if they are in the right place. I aligned them with the urls, is that ok?

michaeljguarino: yeah they should be horizontally aligned w/ the urls

assets/src/components/cd/services/service/ServiceDetailsSidecar.tsx
```diff
@@ -103,6 +139,24 @@ export function ServiceDetailsSidecar({
             {helm.version}
           </SidecarItem>
         )}
+        {repository && (
+          <SidecarItem heading="Git repository">
+            <div
```
michaeljguarino, line 144: Is there not some reusable component for this already?  If there isn't we probably should make one and put it in the utils library, can't imagine we won't need more icon-ed strings in the future.

michaeljguarino (approved):

PR #809, opened 2024-04-04
Pipeline gate force open/close UI

renemennab:
## Summary
Adds buttons to allow user to force the gate state on a pipeline job
The buttons only appear if the gates state is `RUNNING` or `PENDING`

image

it also changes the position of the global services tab

image

[linear task](https://linear.app/pluralsh/issue/ENG-1821/pipeline-gate-force-openclose-ui#comment-68b96bb4)

### DIFF
--- a/assets/src/components/cd/pipelines/job/PipelineJobLogs.tsx
+++ b/assets/src/components/cd/pipelines/job/PipelineJobLogs.tsx
@@ -1,6 +1,17 @@
 import { useMemo, useState } from 'react'
-import { Button, FormField, ListBoxItem, Select } from '@pluralsh/design-system'
-import { useJobGateLogsQuery } from 'generated/graphql'
+import {
+  Banner,
+  Button,
+  FormField,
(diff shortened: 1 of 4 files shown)


assets/src/components/cd/pipelines/job/PipelineJobLogs.tsx
```diff
@@ -66,6 +83,45 @@ export default function PipelineJobLogs() {
     <ScrollablePage
       heading="Logs"
       scrollable={false}
+      headingContent={
+        jobGateState === 'RUNNING' || jobGateState === 'PENDING' ? (
```
michaeljguarino, line 87: shouldn't we be able to get the `RUNNING`, `PENDING` strings from the generated types so they aren't magic values in the code?

renemennab: yes, totally. I'll fix that.

michaeljguarino (approved):

PR #811, opened 2024-04-05
feat(chart): bump kubernetes-dashboard chart dep to latest version

floreks:
## Checklist


templates/values.yaml.tpl
```diff
@@ -59,4 +59,12 @@ kas:
...
     auth:
-      password: {{ .Values.kasRedis }}
\ No newline at end of file
+      password: {{ .Values.kasRedis }}
+
+dashboard:
+  api:
+    containers:
+      args:
+        - --apiserver-host=https://{{ .Values.kasDns }}/k8s-proxy
```
michaeljguarino, line 68: We should also add this to the `templates/values.yaml.liquid`.  I think this is used by plural up the other by `plural cd control-plane` which generates helm values for byok installs

floreks: It is already there. I have added it to the initial PR that introduced dashboard dependency. Just forgot to update this one too.

maciaszczykm (approved):

PR #818, opened 2024-04-08
limit service list to same cluster

renemennab:
## Summary
Limits the service list in the service details page to services within the same cluster

image

### DIFF
--- a/assets/src/components/cd/services/service/ServiceDetails.tsx
+++ b/assets/src/components/cd/services/service/ServiceDetails.tsx
@@ -151,15 +151,13 @@ function ServiceDetailsBase() {
   const logsEnabled = useLogsEnabled()
 
   const { data: serviceListData } = useServiceDeploymentsTinyQuery({
+    variables: { clusterId },
     pollInterval: POLL_INTERVAL,
     fetchPolicy: 'cache-and-network',
   })
   const serviceList = useMemo(
-    () =>
-      mapExistingNodes(serviceListData?.serviceDeployments).filter(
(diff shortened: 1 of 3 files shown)


assets/src/components/cd/services/service/ServiceDetails.tsx
```diff
@@ -155,8 +155,11 @@ function ServiceDetailsBase() {
     fetchPolicy: 'cache-and-network',
   })
   const serviceList = useMemo(
-    () => mapExistingNodes(serviceListData?.serviceDeployments),
-    [serviceListData?.serviceDeployments]
+    () =>
```
michaeljguarino, line 158: this should be done in the query not in a postfilter

assets/src/graph/cdServices.graphql
```diff
@@ -165,8 +165,8 @@ query ServiceDeployments(
   }
 }
 
-query ServiceDeploymentsTiny {
-  serviceDeployments(first: 100) {
+query ServiceDeploymentsTiny($clusterId: ID!) {
```
michaeljguarino, line 168: Is this only used in the selector?  We can type the `$clusterId` arg as `ID` and allow it to still select all services for other usecases if need be (by omitting or null-ing the $clusterId arg).

renemennab: yes, it is only used in the selector, but I agree, allowing it to be null is a good idea. I'll change that.

michaeljguarino (approved):

PR #831, opened 2024-04-11
add create scm webhook button

renemennab:
Adds a create webhook button to the Smc Webhooks page. this button opens a modal

image

After the user creates it the modal shows the url and the secret,

I didn't test the changed state because i didn't know if I could create a new webhook, but since this is a time sensitive issue, i wanted to submit for review asap

### DIFF
--- a/assets/src/components/pr/scm/CreateScmWebhook.tsx
+++ b/assets/src/components/pr/scm/CreateScmWebhook.tsx
@@ -1,5 +1,11 @@
 import { useCallback, useState } from 'react'
-import { Button, FormField, Input2, Modal } from '@pluralsh/design-system'
+import {
+  Button,
+  Codeline,
+  FormField,
+  Input2,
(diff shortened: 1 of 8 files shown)


assets/src/components/pr/scm/CreateScmWebhook.tsx
```diff
@@ -108,75 +100,84 @@ export function CreateScmWebhookModalBase({
...
+            header="Something went wrong"
+            error={error}
           />
-          {error && <GqlError error={error} />}
-        </div>
-      )}
+        )}
+        {newWebHook && !error && (
+          <>
+            <FormField label="Webhook URL">
```
michaeljguarino, line 120: Could we add a bit of text field here that says:

 "Add a new webhook in your source control provider with the following url and validation secret"

michaeljguarino (approved):

PR #849, opened 2024-04-16
docs: generate docs from the CRDs

zreigz:
## Test Plan
generate docs from the CRDs:
```
$ make crd-docs
```

## Checklist


controller/config.yaml
```diff
@@ -0,0 +1,17 @@
+processor:
+  ignoreTypes:
+    - "(Elasticsearch|Kibana|ApmServer)List$"
```
michaeljguarino, line 3: These specific ignore types aren't useful for us I think, we should probably do something like `.*List$` though

controller/config.yaml
```diff
@@ -0,0 +1,17 @@
...
+    - "ElasticsearchSettings$"
+    - "Associa(ted|tor|tionStatus|tionConf)$"
+  ignoreFields:
+    - "status$"
+    - "TypeMeta$"
+
+render:
+  kubernetesVersion: 1.22
+  knownTypes:
+    - name: SecretObjectReference
```
michaeljguarino, line 15: probably should add ObjectReference and some other core types we use

zreigz: done

PR #852, opened 2024-04-17
Update memory queries to ensure only containers are measured

snikch:
Context: https://discord.com/channels/880830238723047424/1229549563678031893


snikch:
ok, I decided to bump it rather than bum it but I'm pretty sure that's what you meant 😂 .

michaeljguarino (approved):

PR #853, opened 2024-04-17
add managed namespaces views

renemennab:
- Removes providers tab in continuous deployment
- ads namespaces tab and namespaces detail view.

Questions:
- should there be a `create namespace` button?
- Are the tables ok? Are those the attributes that should be displayed?
- you requested that the detail view should have a service table, but the namespace request only returns one single service, is that correct?


michaeljguarino:
@renemennab I added an additional `services` connection field to the gql schema that should solve the single service issue.  You were actually fetching the service template not the spawned services.

renemennab:
I fixed the details page to use the proper services, but i could't test it because there are no services being returned in the test namespace.

michaeljguarino (changes requested):
You shouldn't delete all the providers code, just don't show the tab by default.  We'll find another place to visualize it.

michaeljguarino:
> I fixed the details page to use the proper services, but i could't test it because there are no services being returned in the test namespace.

I probably know why that's happening, give me a bit to rewire it, but I doubt there's any frontend change given the queries you provided

michaeljguarino (changes requested):
The scrolling is bugged (I think this is true for global services too actually)

michaeljguarino (approved):

PR #863, opened 2024-04-19
show cluster metadata when available

renemennab:
displays cluster metadata on a read-only code block in the metadata tab


assets/src/components/cd/cluster/ClusterMetadata.tsx
```diff
@@ -47,6 +48,8 @@ function MetadataCard({
   const status = cluster?.status
   const renderTag = (tag) => `${tag.name}${tag.value ? `: ${tag.value}` : ''}`
 
+  console.log('cluster', cluster)
```
michaeljguarino, line 51: rm

assets/src/components/cd/cluster/ClusterMetadata.tsx
```diff
@@ -149,6 +152,11 @@ function MetadataCard({
           />
         </section>
       )}
+      {cluster.metadata && (
+        <Code language="json">
+          {formatJson(JSON.stringify(cluster.metadata))}
```
michaeljguarino, line 157: why do you need to stringify -> parse -> stringify here?  My guess is the only issue you'll face is null checking, and the api should guarantee it's valid json

michaeljguarino (approved):

PR #886, opened 2024-04-25
implement new upgrade plan designs

jsladerman:
Implemented new design for cluster upgrade plans
Made a small change to where deprecation info is fetched to reduce sideloads when listing clusters


assets/src/components/cd/clusters/ClusterUpgradeFlyover.tsx
```diff
@@ -0,0 +1,426 @@
...
+function ClustersUpgradeNow({
+  cluster,
+  targetVersion,
+  apiDeprecations,
+  refetch,
+  setError,
+}: {
+  cluster?: ClustersRowFragment | null
+  targetVersion: Nullable<string>
+  apiDeprecations: unknown[]
```
michaeljguarino, line 96: why is this typed as `unknown[]`?

jsladerman: ah good catch- was a placeholder for the linter. fixed

PR #890, opened 2024-04-26
feat: allow manage dependencies from the API

zreigz:
## Test Plan
Allow API to manage service dependencies 

## Checklist


controller/internal/controller/servicedeployment_controller.go
```diff
@@ -245,6 +235,18 @@ func (r *ServiceReconciler) genServiceAttributes(ctx context.Context, service *v
 		Templated:       lo.ToPtr(true),
 	}
 
+	if len(service.Spec.Dependencies) > 0 {
+		attr.Dependencies = make([]*console.ServiceDependencyAttributes, 0)
+	}
+	for _, dep := range service.Spec.Dependencies {
+		serviceDep := &v1alpha1.ServiceDeployment{}
```
michaeljguarino, line 242: We shouldn't need to do this resolution, it can just store the name straight from the input string.  (The reason being we want to be able to support things like global services that spawn services w/o associated visible crds).

This would need to go for the other genServiceTemplate func too

zreigz: done

michaeljguarino (approved):

PR #892, opened 2024-04-26
display service dependencies in service details

jsladerman:
added table to list name and status of service dependencies


assets/src/components/cd/services/ServiceDependenciesColumns.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { createColumnHelper } from '@tanstack/react-table'
+
+const columnHelper = createColumnHelper<Nullable<ServiceDependency>>()
+
+const ColName = columnHelper.accessor('name', {
+  id: 'name',
+  header: 'Name',
+})
+
+const ColStatus = columnHelper.accessor('status', {
```
michaeljguarino, line 11: I think we should render the status as a chip with appropriate color.  Also, it's realistically only set as either nil or `HEALTHY` at the moment so there'd need to be ternary logic like `color={status === 'HEALTHY' ? 'success' : 'warn'}`

jsladerman: updated

assets/src/components/cd/services/ServiceDependenciesColumns.tsx
```diff
@@ -0,0 +1,27 @@
...
+  id: 'status',
+  header: 'Status',
+  cell: ({ getValue }) => (
+    <div>
```
michaeljguarino, line 16: is the parent div needed?

assets/src/components/cd/services/ServiceDependenciesColumns.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <Chip
+        size="large"
+        severity={getValue() === 'HEALTHY' ? 'success' : 'warning'}
```
michaeljguarino, line 19: ideally we use the actual codegened enum here, but can't remember how to import it

michaeljguarino (approved):

assets/src/components/cd/services/ServiceDependenciesColumns.tsx
jsladerman: yeah good call. typescript was still throwing the warning if the checked string wasn't part of the enum but I went ahead and imported it

PR #899, opened 2024-04-29
Kubernetes dashboard tab renaming

jsladerman:
Renamed and refactored accordingly:
Discovery -> Network
Access -> RBAC


assets/src/components/kubernetes/rbac/RBAC.tsx
```diff
@@ -38,7 +38,7 @@ const directory = [
   { path: SERVICE_ACCOUNTS_REL_PATH, label: 'Service Accounts' },
 ] as const
 
-export default function Access() {
+export default function RBAC() {
```
michaeljguarino, line 41: I'm not sure the convension around casing these acronyms, but i feel like `Rbac` and `Rbac.tsx` seems more correct perhaps?

jsladerman: fixed

michaeljguarino: I think the filename is still `RBAC.tsx` though

jsladerman: Yeah just realized git ignored the casing change. fixed

michaeljguarino (approved):

PR #905, opened 2024-04-30
policy management global view

renemennab:
## Checklist


assets/src/graph/policies.graphql
```diff
@@ -0,0 +1,66 @@
...
+  insertedAt
+  name
+  recommendation
+  ref {
+    kind
+    name
+  }
+  updatedAt
+  violationCount
+  violations {
```
michaeljguarino, line 16: we probably don't want to sideload this on every query of a `PolicyConstraint`.  Ideally we'd only use them on the detail view of a constraint unless it's absolutely necessary.

assets/src/components/policies/PoliciesViolationsGauge.tsx
```diff
@@ -0,0 +1,122 @@
+import { Subtitle2H1 } from 'components/utils/typography/Text'
+import { Flex } from 'honorable'
+import styled, { useTheme } from 'styled-components'
+
+const violationLabels = ['WITHOUT VIOLATIONS', 'WITH VIOLATIONS']
+
+function PoliciesViolationsGauge({
```
michaeljguarino, line 7: we're going to need to use these gauges for a few views, and it'd be ideal to have a reusable solution.  I had used https://nivo.rocks/ (this has some benefits like entry animations too) for some of this in the past, and we can theoretically use recharts as well.  If none are good, we should implement it in the design system and import here so it can be added to other components.

michaeljguarino (approved):

PR #907, opened 2024-05-01
added indicator to service sync button

jsladerman:
Added a line to the tooltip indicating last time a service's git repo was pulled

Sync button also disables with a countdown if it's been < 15 seconds since last pull


assets/src/components/apps/Apps.tsx
```diff
@@ -158,7 +158,7 @@ export default function Apps() {
   const navigate = useNavigate()
 
   useEffect(() => {
-    if (configuration?.byok) navigate('/cd/clusters')
+    if (configuration?.byok) navigate('/home')
```
michaeljguarino, line 161: should revert this for now

michaeljguarino (changes requested):
Looks like some work on the home screen got interleaved here.  That's actually fine, just make sure it's not routable at least

jsladerman:
> Looks like some work on the home screen got interleaved here. That's actually fine, just make sure it's not routable at least

ah looks like I was still on that branch when I created this one. took the home screen commit out should be good now

michaeljguarino (approved):

PR #908, opened 2024-05-01
Add Pod Logs view to k8s dashboard

michaeljguarino:
This is not fully implemented, but functional for now.  Will at least give log views in the k8s dashboard for users who might want them.

## Test Plan
local

## Checklist


jsladerman (approved):
added a small tweak for TS and linter otherwise lgtm

PR #935, opened 2024-05-08
fix: add clusterid param back to service table query

jsladerman:
The param was left out by mistake during a recent refactor


jsladerman:
fixed in #932

PR #949, opened 2024-05-10
Handle byok clusters slightly better

michaeljguarino:
The usecase here is:
* user creates cluster CRD for byok cluster prior to installing agent
* CRD should retry until agent is there then reconcile successfully

## Test Plan
n/a

## Checklist


controller/internal/controller/cluster_controller_test.go
```diff
@@ -295,18 +295,20 @@ var _ = Describe("Cluster Controller", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(sanitizeClusterStatus(cluster.Status)).To(Equal(sanitizeClusterStatus(v1alpha1.ClusterStatus{
 				Status: v1alpha1.Status{
-					ID:  lo.ToPtr(byokClusterConsoleID),
-					SHA: lo.ToPtr("CPYLCGRGF2JWFBF3OGRHQQUSBDXW6Y4VMUDQDCQQDEA6G6CAZORQ===="),
+					// ID:  lo.ToPtr(byokClusterConsoleID),
```
seemywingz, line 298: Could you tell me what the policy is for committing commented code?
Is this for reference or testing?
I think I'm still a bit traumatized from Amazon code reviews 😬

seemywingz (approved):

controller/internal/controller/cluster_controller_test.go
michaeljguarino: There's no real policy, I commented it out because i wasn't sure whether i needed to revert

zreigz (approved):

PR #958, opened 2024-05-14
Add expanded state to sidebar

renemennab:
image


jsladerman:
The icon is in kind of a strange spot, could it be put above the nav bar? Like in this design:
Screenshot 2024-05-14 at 4 03 52 PM
Screenshot 2024-05-14 at 4 04 04 PM

jsladerman (approved):

assets/src/components/layout/Sidebar.tsx
```diff
@@ -77,51 +80,59 @@ function getMenuItems({
...
       text: 'Apps',
+      expandedLabel: 'Apps',
       icon: <AppsIcon />,
       path: '/',
       plural: true,
       pathRegexp: /^\/(apps)/,
     },
     {
       text: 'Continuous deployment',
+      expandedLabel: 'CD',
```
maciaszczykm, line 97: I know it's a little bit different but to keep things simpler I'd use `text` as `expandedLabel` and figure out which variant to use.

PR #959, opened 2024-05-14
optimize cluster and service search bars

jsladerman:
The search inputs/filters are pretty laggy (very noticeable on mid-lower end devices). Debouncing solves the issue of reducing network requests but doesn't prevent rerendering the whole table on every keystroke. I decoupled the input value states (what's displayed) from the query values so basically the table rendering is also debounced or deferred

The other solution to this would be memoizing the tables (and also all of the props passed to them) but that would probably require a much bigger refactor

Also fixed a bug with the cluster status filters not actually filtering

### DIFF
--- a/assets/src/components/cd/services/ServicesTable.tsx
+++ b/assets/src/components/cd/services/ServicesTable.tsx
@@ -4,7 +4,6 @@ import { useNavigate } from 'react-router'
 import { useTheme } from 'styled-components'
 import type { Row } from '@tanstack/react-table'
 import isEmpty from 'lodash/isEmpty'
-import { useDebounce } from '@react-hooks-library/core'
 import {
   ServiceDeploymentStatus,
   type ServiceDeploymentsRowFragment,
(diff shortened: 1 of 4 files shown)


assets/src/components/cd/services/ServicesTable.tsx
```diff
@@ -56,9 +55,9 @@ export function ServicesTable({
       queryKey: 'serviceDeployments',
     },
     {
-      q: debouncedSearchString,
+      q: queryString,
```
maciaszczykm, line 58: Why are you removing debounce here?

jsladerman: It was just moved into the component

maciaszczykm: Oh, okay. Not easy to read code here instead of IDE.

maciaszczykm (approved):

PR #967, opened 2024-05-17
remove info view from kubernetes cluster namespaces

renemennab:


assets/src/components/kubernetes/cluster/Namespace.tsx
```diff
@@ -88,86 +75,6 @@ export default function Namespace(): ReactElement {
...
-    cell: ({ getValue }) => getValue(),
-  }),
-  lrColumnHelper.accessor((lr) => lr?.defaultRequest, {
-    id: 'defaultRequest',
-    header: 'Default request',
-    cell: ({ getValue }) => getValue(),
-  }),
-]
-
-export function NamespaceInfo(): ReactElement {
```
michaeljguarino, line 136: Let's keep this code around for when we might want to rethink the view

michaeljguarino (approved):

PR #970, opened 2024-05-17
add manager view to home screen

jsladerman:
Added a card to the home page with constraint violations chart/table for admins


assets/src/components/home/Home.tsx
```diff
@@ -28,6 +29,7 @@ export default function Home() {
       />
       <HomeContentWrapperSC>
         <ClusterOverviewCard />
+        {me?.roles?.admin && <ConstraintViolationsCard />}
```
michaeljguarino, line 32: This isn't the field we'd make it conditional on.  It'd actually be from the users personas and if one of them specifies the `home` field as `manager`.

jsladerman: Ah yeah that makes more sense. fixed (made a test persona to confirm all seems to work now)

assets/src/components/home/Home.tsx
```diff
@@ -13,13 +13,19 @@ import { ClusterOverviewCard } from './clusteroverview/ClusterOverviewCard'
...
 import { PrCard } from './pullrequests/PrCard'
+import { ConstraintViolationsCard } from './managerview/violations/ConstraintViolationsCard'
 
 const breadcrumbs: Breadcrumb[] = [{ label: 'home', url: '/' }]
 
 export default function Home() {
   useSetBreadcrumbs(breadcrumbs)
   const { me } = useContext(LoginContext)
 
+  const isManager =
```
michaeljguarino, line 24: we should probably make some hooks for this (in case they'll be needed to reconfigure other ui elements)

michaeljguarino (approved):

PR #973, opened 2024-05-22
feat: added keda and zalando-postgres

smyja:
Added Keda, a graduated CNCF project and Zalando postgres operator. 

Keda: https://keda.sh/docs/2.14/operate/cluster/#kubernetes-compatibility
Postgres Zalando: https://github.com/zalando/postgres-operator#supported-postgres--k8s-versions
 

## Checklist

### DIFF
--- a/static/compatibilities/manifest.yaml
+++ b/static/compatibilities/manifest.yaml
@@ -15,3 +15,5 @@ names:
 - kyverno
 - vitess
 - strimzi
+- keda
+- zalando-postgres
--- a/static/compatibilities/manifest.yaml
+++ b/static/compatibilities/manifest.yaml
@@ -16,4 +16,4 @@ names:
 - vitess
 - strimzi
 - keda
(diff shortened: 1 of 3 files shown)


static/compatibilities/manifest.yaml
```diff
@@ -14,4 +14,6 @@ names:
 - opentelemetry
 - kyverno
 - vitess
-- strimzi
\ No newline at end of file
+- strimzi
+- keda
+- zalando-postgres
```
michaeljguarino, line 19: this name is not going to actually work, I'm not positive what it should be but my guess is `postgres-operator`

smyja: > this name is not going to actually work, I'm not positive what it should be but my guess is `postgres-operator`

I've changed it to postgres-operator.

smyja:
@michaeljguarino PTAL

PR #987, opened 2024-05-27
Use PENDING_APPROVAL state to control approve displays

michaeljguarino:
At the moment it shows a lot of these prematurely, having harness set this state appropriately solves for that.

## Test Plan
n/a

## Checklist

### DIFF
--- a/assets/src/components/stacks/run/Sidecar.tsx
+++ b/assets/src/components/stacks/run/Sidecar.tsx
@@ -68,14 +68,15 @@ export default function StackRunSidecar({
       flexDirection="column"
       gap="small"
     >
-      {stackRun.approval && !stackRun.approvedAt && (
-        <Button
-          onClick={mutation}
-          loading={loading}
-        >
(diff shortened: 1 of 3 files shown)


assets/src/components/stacks/run/Sidecar.tsx
```diff
@@ -68,7 +68,7 @@ export default function StackRunSidecar({
       flexDirection="column"
       gap="small"
     >
-      {stackRun.approval && !stackRun.approvedAt && (
+      {stackRun.status == StackStatus.PendingApproval && !stackRun.approvedAt && (
```
jsladerman, line 71: Nit but should be ===

jsladerman (approved):

assets/src/components/stacks/run/Sidecar.tsx
jsladerman: Also 'yarn fix' should fix the lint error if it's preventing a merge

PR #1007, opened 2024-06-03
fix: Few fixes for stack and stack run job views

maciaszczykm:
Mostly styling/routing changes.


michaeljguarino (approved):
think there might be a linter issue though

PR #1013, opened 2024-06-04
fix: handle delete resource error better

floreks:
- Properly display delete error response in the delete resource dialog
- Improve k8s dashboard zerostate handling when there are no clusters

image

## Test Plan

Locally

## Checklist

### DIFF
--- a/assets/src/components/kubernetes/common/DeleteResource.tsx
+++ b/assets/src/components/kubernetes/common/DeleteResource.tsx
@@ -10,6 +10,7 @@ import {
 import { useParams } from 'react-router-dom'
 import { QueryHookOptions } from '@apollo/client/react/types/types'
 import { useTheme } from 'styled-components'
+import { ApolloError, FetchResult, ServerError } from '@apollo/client'
 
 import { Confirm } from '../../utils/Confirm'
 import {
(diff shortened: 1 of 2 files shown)


assets/src/components/kubernetes/common/DeleteResource.tsx
```diff
@@ -60,10 +61,34 @@ enum DeletionPropagation {
...
+    statusCode: networkError?.statusCode,
+    result: networkError?.result ?? networkError?.message,
+  } as ServerError
+}
+
 function DeleteResourceModal({ open, setOpen, resource, refetch }): ReactNode {
   const theme = useTheme()
   const [deleting, setDeleting] = useState(false)
   const [deleteNow, setDeleteNow] = useState(false)
+  const [serverError, setServerError] = useState<ServerError>()
```
michaeljguarino, line 91: why can't you use the error field of the mutation hook for this?

floreks: That's because the actual error message is part of the response and the error provided directly by apollo does not have the response message. I think that's due to how apollo rest link translates rest responses into GQL. The only place where an error response is available is inside the response `data` returned by the mutation under `errors` field. It took me quite some time to figure out where the error response is stored and how to read it.

michaeljguarino: Oh very weird

floreks: Unfortunately the `data` returned by the `useMutation` hook is different than the one returned by the mutation promise returned when calling the mutation.

Normally I would get it like that but this `data` is undefined when there is an error when calling the mutation, and `error` does not contain the REST request response.
```
const [deleteResource, { error, data }] = useMutation(...)
```

The `data` that I ended up using is returned by the mutation function promise
```
deleteResource().then((data) => setServerError(toServerError(data)))
```

michaeljguarino: whoa ok that's also weird

michaeljguarino (approved):

assets/src/components/kubernetes/common/DeleteResource.tsx
floreks: That is probably due to how apollo rest link plugin handles response transformation and error handling. I haven't seen any better way that would allow us to handle it in a generic way.

PR #1024, opened 2024-06-06
upgrade plan visual improvements

jsladerman:
UX tweaks mostly on the upgrade plan flyover reflecting the new designs

### DIFF
--- a/assets/src/components/cd/clusters/ClusterUpgradeFlyover.tsx
+++ b/assets/src/components/cd/clusters/ClusterUpgradeFlyover.tsx
@@ -2,11 +2,11 @@ import {
   Accordion,
   AppIcon,
   Button,
+  ChecklistIcon,
   ClusterIcon,
   ConfettiIcon,
   Flyover,
   ListBoxItem,
-  MarketPlusIcon,
   Select,
   SuccessIcon,
   Table,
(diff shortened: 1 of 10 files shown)


jsladerman:
(need to merge https://github.com/pluralsh/design-system/pull/603 before this)

assets/src/components/cd/clusters/runtime/columns.tsx
```diff
@@ -166,19 +166,23 @@ const colKubVersion = columnHelperRuntime.accessor((row) => row?.addonVersion, {
...
 
     if (!addonVersion?.blocking) return null
 
     return (
-      <Tooltip label="Blocking">
+      <Chip severity="danger">
         <BlockedIcon
           color="icon-danger"
-          size={16}
+          size="12"
```
michaeljguarino, line 180: are these size values not in the theme?  We should use the theme instead of magic numbers as much as possible

assets/src/components/cd/clusters/ClusterUpgradeFlyover.tsx
```diff
@@ -461,7 +467,10 @@ function ClusterUpgradeAccordionTrigger({
               color="icon-success"
             />
           ) : (
-            <Icon width={16} />
+            <Icon
+              color="icon-warning"
+              width={16}
```
michaeljguarino, line 472: same, should be a theme value

assets/src/components/cd/clusters/runtime/columns.tsx
jsladerman: Yeah I was debating that. The theme has values for spacing which could still be used but not size, felt weird semantically

assets/src/components/cd/clusters/ClusterUpgradeFlyover.tsx
maciaszczykm: IIRC 16px is default value, so it might be possible to skip it.

michaeljguarino (approved):

PR #1026, opened 2024-06-06
add DeploymentSettings CRD

zreigz:
Add DeploymentSettings CRD + operator

## Checklist


controller/api/v1alpha1/deploymentsettings_types.go
```diff
@@ -0,0 +1,113 @@
...
+	// Create bindings.
+	// +kubebuilder:validation:Optional
+	Create []Binding `json:"create,omitempty"`
+
+	// Git bindings.
+	// +kubebuilder:validation:Optional
+	Git []Binding `json:"git,omitempty"`
+}
+
+type StackSettings struct {
```
michaeljguarino, line 80: We should also add `ConnectionRef` here mapping to `connectionId` in the stack settings

controller/internal/controller/deploymentsettings_controller.go
```diff
@@ -0,0 +1,189 @@
...
+func (r *DeploymentSettingsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
+	logger := log.FromContext(ctx)
+
+	settings := &v1alpha1.DeploymentSettings{}
+	if err := r.Get(ctx, req.NamespacedName, settings); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// make sure there is only one CRD object with the `global` name and the agent namespace
+	if settings.Name != deploymentSettingsName || settings.Namespace != deploymentSettingsNamespace {
```
zreigz, line 67: @michaeljguarino I've added such a check to make sure we have only one DeploumentSettings CRD object
with the `global` name and the agent namespace `plrl-deploy-operator`.

controller/api/v1alpha1/deploymentsettings_types.go
zreigz: done

controller/internal/controller/deploymentsettings_controller.go
michaeljguarino: Ok cool that's probably wise

PR #1038, opened 2024-06-08
Add projects to organize clusters/stacks/pipelines

michaeljguarino:
Any of the above can be parented by a project, which has its own permission bindings.  This makes it easier to bulk manage permissions for subsets of resources and also reduces the sprawl in the UI for large dataset sizes. There is a migration concern, which is done by:

* creating a guaranteed unique default project w/in the schema migration
* rewiring all existing clusters/stacks/pipelines to use the default project
* defaulting `project_id` for all create requests to the default project if not specified (old clients thus still work)

Global services and managed namespace apis also are rewired to support project id targets.

## Test Plan
unit test

## Checklist


maciaszczykm (approved):

maciaszczykm:
I'll merge it to be able to work on the frontend and controller implementation.

PR #1039, opened 2024-06-09
change unstructured component default path

jsladerman:
made raw yaml the default when there's not much info

also fixed a routing bug when navigating from service -> component details

### DIFF
--- a/assets/src/components/cd/services/service/ServiceComponents.tsx
+++ b/assets/src/components/cd/services/service/ServiceComponents.tsx
@@ -24,6 +24,8 @@ import { useComponentKindSelect } from 'components/apps/app/components/Component
 import { ComponentList } from 'components/apps/app/components/ComponentList'
 import { ModalMountTransition } from 'components/utils/ModalMountTransition'
 
+import { isUnstructured } from 'components/component/ComponentInfo'
+
 import {
   getServiceDetailsBreadcrumbs,
   useServiceContext,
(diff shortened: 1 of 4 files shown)


assets/src/components/cd/services/service/ServiceComponents.tsx
```diff
@@ -128,6 +128,16 @@ export default function ServiceComponents() {
               ? `${getServiceComponentPath({
                   clusterId,
                   serviceId,
+                  defaultRaw: ![
```
michaeljguarino, line 131: there are definitely some other components that shouldn't default to raw here

assets/src/components/component/ComponentInfo.tsx
```diff
@@ -31,31 +31,25 @@ function hasPods(kind: string): boolean {
...
-    case 'daemonset':
-      return <DaemonSet />
-    case 'canary':
-      return <CanaryInfo />
-    case 'servicedeployment':
-      return <PluralServiceDeployment />
-    default:
-      return undefined
-  }
+  return componentInfoMap[kind]
```
michaeljguarino, line 48: Did you test this with the other component views?  I'd be somewhat concerned it behaves weirdly given that contexts play such a huge role in those components

jsladerman: Yeah I did, didn't find any issues. Context shouldn't be affected because the component tree hasn't changed

michaeljguarino (approved):

PR #1043, opened 2024-06-10
fix: ensure that tokenSecretRef is set before getting secret

floreks:
- Added unit tests to cover basic scenarios

## Test Plan

Locally + unit tests

## Checklist


michaeljguarino (commented):
will this requeue if the scm connection is not found and the tokenSecretRef also is not set?

floreks:
> will this requeue if the scm connection is not found and the tokenSecretRef also is not set?

Since `tokenSecretRef` is optional, it will ignore it and create a ScmConnection in the console API without it.

michaeljguarino:
> > will this requeue if the scm connection is not found and the tokenSecretRef also is not set?
> 
> Since `tokenSecretRef` is optional, it will ignore it and create a ScmConnection in the console API without it.

That's probably not the correct behavior, it should only try to create a ScmConnection if there is a token secret ref (a token is required), but assume if there isn't that it's going to be created outside the operator and requeue until it's created manually by the user.

floreks:
@michaeljguarino I have updated the logic and added a unit test to cover this case.

michaeljguarino (approved):

PR #1046, opened 2024-06-10
link pods tables to k8s dashboard

jsladerman:
changed the pods tables for pipelines, stacks, and service components to point to the k8s view instead of the old one


assets/src/components/cluster/pods/PodsList.tsx
```diff
@@ -420,7 +424,11 @@ export const PodsList = memo(
           reactTableOptions={reactTableOptions}
           {...props}
           onRowClick={(_e, { original }: Row<PodTableRow>) =>
-            navigate(`${linkBasePath}/${original.namespace}/${original.name}`)
+            navigate(
+              `${
+                clusterId ? `${getKubernetesAbsPath(clusterId)}` : ''
```
michaeljguarino, line 429: this is prone to being brittle and breaking if we change these path schemes a bit, it's probably better to just provide the full path for the k8s dashboard and full path for pod views based on clusterId being present

jsladerman: fixed

michaeljguarino (approved):

PR #1049, opened 2024-06-11
Automate compatibility table info

seemywingz:
Add Python Utility to update the compatibility matricoes for tools listed in `static/compatibilities/manifest.yaml`
currently added scrapers and updated compatibility yaml:
- ingress-nginx
- cert-manager
- external-dns

## Test Plan

tested manually, unit test will be added

## Checklist


static/compatibilities/cert-manager.yaml
```diff
@@ -2,6 +2,10 @@ icon: https://raw.githubusercontent.com/cert-manager/cert-manager/d53c0b9270f8cd
 git_url: https://github.com/cert-manager/cert-manager
 release_url: https://github.com/cert-manager/cert-manager/releases/tag/v{vsn}
 versions:
+- version: 1.15.0
+  kube: ['1.24', '1.25', '1.26', '1.27', '1.28', '1.29', '1.30']
```
michaeljguarino, line 6: why is this one in ascending order while others are descending?

michaeljguarino (commented):
Generally want to sort versions in ascending order because the latest version is most significant info

utils/compatibility/scrapers/ingress-nginx.py
```diff
@@ -0,0 +1,52 @@
+# scrapers/ingress_nginx.py
+
+import requests
+from bs4 import BeautifulSoup
+from collections import OrderedDict
+from utils import printError, update_compatibility_info
+
+
+def scrape():
```
michaeljguarino, line 9: This is pretty classic "arrow-code", putting a whole lot of logic in nested if-statements that becomes hard to maintain.  The simple solve is to invert the logic, eg use stuff like:

```python
if response.status_code != 200:
  print_error("could not fetch url")
  return

...rest of logic below
```

Pretty sure the entire function can be re-organized along those lines and require at most one indentation level.

static/compatibilities/cert-manager.yaml
seemywingz: Hmm, I'm not sure; that's definitely weird. I'll look into it.

utils/compatibility/scrapers/ingress-nginx.py
seemywingz: Makes sense, I'll refactor this and the other scrapers

static/compatibilities/cert-manager.yaml
```diff
@@ -2,20 +2,24 @@ icon: https://raw.githubusercontent.com/cert-manager/cert-manager/d53c0b9270f8cd
...
+  requirements: []
+  incompatibilities: []
 - version: 1.14.0
-  kube: ['1.29', '1.28', '1.27', '1.26', '1.25', '1.24', '1.24']
+  kube: ['1.30', '1.29', '1.28', '1.27', '1.26', '1.25', '1.24']
   requirements: []
   incompatibilities: []
 - version: 1.13.0
-  kube: ['1.28', '1.27', '1.26', '1.25', '1.24', '1.23']
+  kube: ['1.27', '1.26', '1.25', '1.24', '1.23', '1.22', '1.21']
```
michaeljguarino, line 14: This has to be wrong if 1.12 is compatible up to 1.30 right?

utils/compatibility/main.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+manifestFile = "../../static/compatibilities/manifest.yaml"
+
+if not os.path.exists(manifestFile):
+    printError(f"Manifest file not found at {manifestFile}")
+else:
+    manifest = readYaml(manifestFile)
+    print()
+    if manifest:
```
michaeljguarino, line 27: this is still arrow-code

utils/compatibility/utils.py
```diff
@@ -0,0 +1,111 @@
+import yaml
+import requests
+
+from collections import OrderedDict
+from colorama import Fore, Style
+from packaging.version import Version
+
+
+def printError(message):
```
michaeljguarino, line 9: we should standardize on snake casing functions since that is the python standard

michaeljguarino (approved):

PR #1072, opened 2024-06-17
added logic for nested data to slice polling

jsladerman:
added polling to global services and managed namespace services tables

they originally weren't able to use the same logic for polling as other tables because the paginated data isn't at the top level for these queries (e.g. it's under `{globalService: {services: {paginatedData} }` as opposed to most other tables which are just `{keyName: {paginatedData} }`). made an attempt to generalize this case on top of the existing logic. seems to be working now, though we might want to clean this up at some point

### DIFF
--- a/assets/src/components/cd/utils/useFetchPaginatedData.tsx
+++ b/assets/src/components/cd/utils/useFetchPaginatedData.tsx
@@ -1,7 +1,10 @@
-import { useCallback, useState } from 'react'
+import { useCallback, useMemo, useState } from 'react'
 import { VirtualItem } from '@tanstack/react-virtual'
-import { extendConnection } from 'utils/graphql'
-import { useSlicePolling } from 'components/utils/tableFetchHelpers'
+import { extendConnection, updateNestedConnection } from 'utils/graphql'
+import {
+  reduceNestedData,
+  useSlicePolling,
+} from 'components/utils/tableFetchHelpers'
 import { POLL_INTERVAL } from 'components/cd/ContinuousDeployment'
 import {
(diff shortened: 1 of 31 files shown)


assets/src/utils/graphql.ts
```diff
@@ -44,6 +44,27 @@ export function extendConnection<
   }
 }
 
+// for cases where paginated data isn't at the top level of the query
+export function updateNestedConnection<TData>(
+  nestedPath: string[],
+  fullQuery: TData,
+  newConnection: any
+): TData {
+  const res = { ...fullQuery }
```
michaeljguarino, line 53: Why do you need to reconstruct the object here?

assets/src/components/cd/utils/useFetchPaginatedData.tsx
```diff
@@ -21,9 +24,10 @@ type GenericQueryHook<TQueryType, TVariables extends OperationVariables> = (
 type FetchDataOptions<TQueryType, TVariables extends OperationVariables> = {
   queryHook: GenericQueryHook<TQueryType, TVariables>
   pageSize?: number
-  queryKey: string
+  queryKey: string // this should be the last field of the nested path if one is provided
```
michaeljguarino, line 27: maybe don't require this and just infer it from `nestedPath`

assets/src/utils/graphql.ts
michaeljguarino: oh wait nvm i get it now

michaeljguarino (approved):

PR #1077, opened 2024-06-17
build: Prod 2278/use a cron GitHub action to automate compatibility gathering

seemywingz:
This adds a GitHub Workflow Cron Job that triggers every day at midnight and runs the compatibility scraper then submits a PR with the updates

## Test Plan

I tested the workflow manually within GitHub Actions

## Checklist


.github/workflows/compatibility-updater.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - name: Install dependencies
+      run: |
+        # Install pyenv dependencies
+        sudo apt-get update
+        sudo apt-get install -y make build-essential libssl-dev zlib1g-dev \
+            libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm \
+            libncurses5-dev libncursesw5-dev xz-utils tk-dev libffi-dev \
+            liblzma-dev
+
+    - name: Install pyenv
```
michaeljguarino, line 28: could we not use something like: https://github.com/marketplace/actions/pyenv-build

.github/workflows/compatibility-updater.yaml
```diff
@@ -0,0 +1,61 @@
...
+        eval "$(pyenv init -)"
+        cd $SCRIPTS_DIR
+        source setup_env.sh
+        python main.py    
+
+    - name: Get current date
+      id: date
+      run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+
+    - name: Commit Changes
```
michaeljguarino, line 46: the create-pull-request action should commit for you

.github/workflows/compatibility-updater.yaml
seemywingz: We could, but it doesn't necessarily improve the runtime speed or flow.
I've scripted the Python env setup to be dynamic and repeatable locally and in the GitHub action. 
Doing it via the script simplifies the maintenance of local dev and won't require any changes to the workflow yaml

michaeljguarino: oh no it just makes it easier to set up pyenv in the github action i would expect, instead of manually doing it ourselves.

.github/workflows/compatibility-updater.yaml
seemywingz: I removed the `Commit Changes` step and tested, it's working as expected

.github/workflows/compatibility-updater.yaml
seemywingz: Would you prefer we use that action?
I can make the changes and test using it.

seemywingz: I made the changes to use the Pyenv Builder Action and reopened the PR

seemywingz: It's using an older version of pyenv, so I had to downgrade the python version from `3.12.3` to `3.12.0`, everything tested fine though

michaeljguarino (approved):

PR #1078, opened 2024-06-17
feat: added harbor compatibility

smyja:
## Checklist


michaeljguarino (commented):
harbor is a major cncf project but isn't properly a k8s add-on (will just be deployed to a single tools cluster usually), so isn't quite in scope for this feature in our minds

PR #1083, opened 2024-06-18
feat: Add imports to ServiceDeployment CRD

maciaszczykm:


controller/api/v1alpha1/servicedeployment_types.go
```diff
@@ -109,12 +158,35 @@ type ServiceSpec struct {
 	// Templated should apply liquid templating to raw yaml files, defaults to true
 	// +kubebuilder:validation:Optional
 	Templated *bool `json:"templated,omitempty"`
-
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Imports are immutable"
+	Imports []corev1.ObjectReference `json:"imports"`
```
michaeljguarino, line 163: I think to make this more future proof, we should have a type here like:

```
- stackRef:
      name: first-stack    
- stackRef:
      name: other-stack
```

That way we can add adjacent fields w/o breaking the schema

maciaszczykm: Done.

michaeljguarino (approved):
Are we adding owner references to these resources?  Without requeueing we might not be able to gather changes on them easily

maciaszczykm:
I'll add it.

maciaszczykm:
Added and verified that it is set.

zreigz (approved):

PR #1099, opened 2024-06-21
fix: allow resource creation when group/user bindings not found

floreks:
## Checklist


michaeljguarino (commented):
what happens when this gets added as attributes?  Maybe we should filter if both groupId and userId is null

floreks:
If group/user is not found then it will return an empty string as `id`. `lo.EmptyableToPtr` will return `nil` if zero value is provided, so the attributes will not change. Only if `id` is non-empty, it will update attributes.

michaeljguarino (approved):

PR #1125, opened 2024-07-01
fix: properly cache-and-network on k8s dashboard resource lists

floreks:
Fix for fetch policy on k8s dashboard resource list views. It will now get data from the cache if available to show immediately and at the same time make an API request to get the latest data.

## Test Plan

Locally

## Checklist

### DIFF
--- a/assets/src/components/kubernetes/common/ResourceList.tsx
+++ b/assets/src/components/kubernetes/common/ResourceList.tsx
@@ -115,6 +115,7 @@ export function ResourceList<
     client: KubernetesClient(cluster?.id ?? ''),
     skip: !cluster,
     pollInterval: 30_000,
+    fetchPolicy: 'cache-and-network',
     variables: {
       filterBy: `name,${filter}`,
       sortBy,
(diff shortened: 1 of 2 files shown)


jsladerman (approved):

michaeljguarino (approved):

michaeljguarino (commented):
Perhaps the poll interval should be dropped to 10s too?

floreks:
10s could not work properly on i.e. pod list on a cluster with lots of pods 1k+. The time to fetch the data could be longer than the refresh interval. Unless we add a caching layer to dashboard API, it might cause issues.

michaeljguarino:
> 10s could not work properly on i.e. pod list on a cluster with lots of pods 1k+. The time to fetch the data could be longer than the refresh interval. Unless we add a caching layer to dashboard API, it might cause issues.

That seems relatively fine, does apollo not have internal logic to ensure polls don't overlap though?  I suppose it could still cause too much load on the underlying pod.

PR #1126, opened 2024-07-01
feat: Add `ServiceAccount` CRD

maciaszczykm:
## Checklist


controller/internal/client/serviceaccount.go
```diff
@@ -0,0 +1,53 @@
...
+func (c *client) UpdateServiceAccount(ctx context.Context, id string, attributes console.ServiceAccountAttributes) (*console.UserFragment, error) {
+	response, err := c.consoleClient.UpdateServiceAccount(ctx, id, attributes)
+	if err != nil {
+		return nil, err
+	}
+
+	return response.UpdateServiceAccount, err
+}
+
+func (c *client) DeleteServiceAccount(ctx context.Context, id string) error {
```
michaeljguarino, line 37: I don't think we need to delete the service account actually (and it might not be strictly safe to do so)

maciaszczykm: Ok, I have removed that part.

controller/internal/controller/service_account_controller.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		logger.Error(err, "unable to create or update sa")
+		utils.MarkCondition(sa.SetCondition, v1alpha1.SynchronizedConditionType, v1.ConditionFalse, v1alpha1.SynchronizedConditionReasonError, err.Error())
+		return ctrl.Result{}, err
+	}
+
+	sa.Status.ID = &apiServiceAccount.ID
+	sa.Status.SHA = &sha
+
+	err = r.syncToken(ctx, sa)
```
michaeljguarino, line 111: will this logic create a new access token on each reconcile?

maciaszczykm: No, it checks if `ServiceAccount` has `ReadyTokenCondition` set to `False` and if that is the case then it exits early from the `syncToken` method. It will only execute if something changed (see lines 96-98).

michaeljguarino: Ah ok

PR #1131, opened 2024-07-02
feat: Operator multitenancy support

maciaszczykm:
Added support in all namespace scoped resource controllers (cluster, cluster restore, cluster restore trigger, deployment settings, service, global service, managed namespace, pipeline, pipeline context, stack, custom stack run, PR automation triggers, notification routers and sinks).

Tried to test multiple scenarios like valid credentials, credentials without access rights, not existing credential secret, missing secret ref. Also checked cache initialisation and flow in debug mode.

Zrzut ekranu 2024-07-4 o 13 19 02
Zrzut ekranu 2024-07-4 o 13 20 38
Zrzut ekranu 2024-07-4 o 12 44 32


controller/internal/controller/servicedeployment_controller.go
```diff
@@ -48,6 +50,11 @@ type ServiceReconciler struct {
 
 func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	logger := log.FromContext(ctx)
+	if err := r.ConsoleClient.UseNamespaceCredentials(req.Namespace, r.CredentialsCache); err != nil {
```
michaeljguarino, line 53: This is going to have severe race condition issues since the console client is a singleton throughout the entirety of this process.  if there are multiple reconcilers (or even one reconciler run in multiple goroutines) executing simultaneously, they could end up using the wrong namespace's credentials.

I think the safest way is to refactor that to return a `client, err` tuple and then use the client in all following code in the individual reconciler from there (the `ServiceReconciler` in this case).  That way the instance is goroutine-local and no race conditions are a concern.

michaeljguarino: I'm not sure it's actually possible, but if you could keep the console client on the reconcilers passed `ctx` that also might be a way to store the global state, or could think of some other mechanic too, it's unfortunate since a ton of code will need to be refactored with a client object as a return value.

michaeljguarino:
This looks great, did you test it for namespaces w/o an attached NamespaceCredentials too to ensure no regressions?

michaeljguarino (approved):
lgtm just want to make sure we're adequately tested just to cover all our bases

maciaszczykm:
> This looks great, did you test it for namespaces w/o an attached NamespaceCredentials too to ensure no regressions?

Yes, I did.

PR #1133, opened 2024-07-03
enhancement: Prod 2310/backfill chart versions on compatibility tables

seemywingz:
Updated the `ingress-nginx` scraper to support new readme for nginx and k8s-supported versions
Add helm chart version updater

## Test Plan

I tested manually and verified the chart versions match the app versions in both the original hem index.yaml and the updated compatibility tables

## Checklist

- [X] If required, I have updated the Plural documentation accordingly.
- [X] I have added a meaningful title and summary to convey the impact of this PR to a user.

### DIFF
--- a/static/compatibilities/cert-manager.yaml
+++ b/static/compatibilities/cert-manager.yaml
@@ -1,6 +1,7 @@
 icon: https://raw.githubusercontent.com/cert-manager/cert-manager/d53c0b9270f8cd90d908460d69502694e1838f5f/logo/logo-small.png
 git_url: https://github.com/cert-manager/cert-manager
 release_url: https://github.com/cert-manager/cert-manager/releases/tag/v{vsn}
+chart_url: https://charts.jetstack.io
 versions:
 - version: 1.15.0
   kube: ['1.30', '1.29', '1.28', '1.27', '1.26', '1.25', '1.24']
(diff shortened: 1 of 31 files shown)


static/compatibilities/cert-manager.yaml
```diff
@@ -1,99 +1,122 @@
 icon: https://raw.githubusercontent.com/cert-manager/cert-manager/d53c0b9270f8cd90d908460d69502694e1838f5f/logo/logo-small.png
 git_url: https://github.com/cert-manager/cert-manager
 release_url: https://github.com/cert-manager/cert-manager/releases/tag/v{vsn}
+chart_url: https://charts.jetstack.io
```
michaeljguarino, line 4: this should probably be `helm_repository_url`

static/compatibilities/ingress-nginx.yaml
```diff
@@ -1,15 +1,18 @@
 icon: https://github.com/pluralsh/plural-artifacts/blob/main/ingress-nginx/plural/icons/nginx.png?raw=true
 git_url: https://github.com/kubernetes/ingress-nginx
 release_url: https://github.com/kubernetes/ingress-nginx/releases/tag/controller-v{vsn}
+chart_url: https://helm.nginx.com/stable
```
michaeljguarino, line 4: this isn't the right helm repository, it's for `nginx` enterprise version of the ingress controller not `ingress-nginx` which is the kubernetes project maintained one.

seemywingz: aaah, right. I developed the chart updated using `cert-manager` and was focused on fixing the `ingress-nginx` scraper when I added the URL. it should be right now.

michaeljguarino (approved):
There are obviously a lot more of these w/ helm repos we should backfill but this is a good start

PR #1144, opened 2024-07-05
light mode fixes

jsladerman:
updated DS and made a couple small tweaks accordingly

### DIFF
--- a/assets/index.html
+++ b/assets/index.html
@@ -1,165 +1,164 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <script>
-      const key = 'theme-mode'
-      function setThemeMode() {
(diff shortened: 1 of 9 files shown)


assets/index.html
```diff
@@ -1,165 +1,164 @@
...
+  </style>
+  <style>
+    html,
+    body {
+      background-color: #171a21;
+    }
+
+    html[data-theme-mode='light'],
+    html[data-theme-mode='light'] body {
+      background-color: #FFFFFF;
```
jsladerman, line 158: this is the only line that actually changed, the rest is due to the formatter but won't happen in future commits

michaeljguarino (approved):

PR #1151, opened 2024-07-07
Delete project via operator

michaeljguarino:
The api supports this now for fully emptied projects, so integrate it w/ the operator as well.

## Test Plan
e2e

## Checklist

### DIFF
--- a/controller/internal/controller/namespace_credentials_controller.go
+++ b/controller/internal/controller/namespace_credentials_controller.go
@@ -74,7 +74,7 @@ func (r *NamespaceCredentialsReconciler) Reconcile(ctx context.Context, req reco
 	// Try to add namespace credentials to cache.
 	if err := r.CredentialsCache.AddNamespaceCredentials(nc); err != nil {
 		utils.MarkFalse(nc.SetCondition, v1alpha1.SynchronizedConditionType, v1alpha1.SynchronizedConditionReasonError, err.Error())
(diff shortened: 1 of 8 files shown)


controller/internal/controller/namespace_credentials_controller.go
```diff
@@ -74,7 +74,7 @@ func (r *NamespaceCredentialsReconciler) Reconcile(ctx context.Context, req reco
 	// Try to add namespace credentials to cache.
 	if err := r.CredentialsCache.AddNamespaceCredentials(nc); err != nil {
 		utils.MarkFalse(nc.SetCondition, v1alpha1.SynchronizedConditionType, v1alpha1.SynchronizedConditionReasonError, err.Error())
-		return ctrl.Result{}, err
+		return requeue, nil
```
maciaszczykm, line 77: Is it done to avoid errors in the logs and keep them only in condition?

maciaszczykm (approved):
LGTM. I have one question.

controller/internal/controller/namespace_credentials_controller.go
michaeljguarino: It's actually more because of the rate limiting controller-runtime imposes when errors are returned.  If it raced, it would potentially throttle to a really long resync period, and the races happen a lot when the service account is created simultaneous to the namespace credentials.

PR #1164, opened 2024-07-11
settings redesign

jsladerman:
consolidated accounts and settings both into the settings tab, implemented new design for it

added CRUD and settings for projects

added page for service accounts


nickcernera:
Looks great overall, nice work.

I only have one note — top bar (title, actions, divider line) and sidecar jump around the screen a little bit when switching tabs. If every view had a title of 40px height, with a 24px gap between it and the divider, then everything should line up.

Not blocking, just some slight VQA.

michaeljguarino:
There's a docker image build failure that needs to be fixed before merged

jsladerman:
> There's a docker image build failure that needs to be fixed before merged

Strange- the errors seem unrelated to any of my changes. Could be from the DS version bump. Will investigate

michaeljguarino (approved):

PR #1171, opened 2024-07-12
Outright digest the git cache tarfile for digest api

michaeljguarino:
We were previously using the commit sha, which can technically miss if there was a commit in a different subfolder.  Just sha-ing the tarball will give a really stable digest for each git cache item.

## Test Plan
n/a

## Checklist


controller/internal/controller/cluster_controller.go
```diff
@@ -127,6 +128,11 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request
 
 	// Sync resource with Console API.
 	apiCluster, err := r.sync(ctx, cluster, providerId, projectId, sha)
+	if err == operrors.ErrRetriable {
```
floreks, line 131: It's safer to use `errors.Is(err, operrors.ErrRetriable)`. In case `err` is wrapped with some other error, it will unwrap all errors in the chain and compare them to `ErrRetriable`.

jsladerman (approved):

PR #1180, opened 2024-07-18
enhancement: React Flow styling changes and refactoring

maciaszczykm:
It's a first step to reduce code duplication in views using React Flow.


jsladerman (approved):
a couple lint errors to fix before merging but otherwise lgtm

PR #1187, opened 2024-07-19
feat: Services tree diagram

maciaszczykm:
Added services tree diagram to both services views. It displays connection between all services. All orphaned nodes are positioned in rows of three elements with the help of invisible edges.

Edit: Fixes PROD-2452.

Zrzut ekranu 2024-07-23 o 14 17 05
Zrzut ekranu 2024-07-23 o 14 18 13


michaeljguarino:
@maciaszczykm Did you also account for the `parent` field on global services (this should also be able to link to the service that spawned the global service)

michaeljguarino:
Also I definitely think this could use a full-screen view like we have for stacks

maciaszczykm:
@michaeljguarino Added missing things and improvements. Now, It also fixes PROD-2452.

michaeljguarino:
Cool cool, I'll give it another look

michaeljguarino (approved):
Looks great, let's merge it

PR #1189, opened 2024-07-19
fix scm connection updates

jsladerman:
was still sending a default struct for github attributes when using token auth


michaeljguarino (approved):
this lgtm but did you run a test (can just create a dummy scm connection then update it then delete it)

jsladerman:
> this lgtm but did you run a test (can just create a dummy scm connection then update it then delete it)

Yeah I did, created one and updated a few different ways toggling between the auth types. All seemed to work right (only sent either the token or GitHub attributes but not both)

PR #1232, opened 2024-07-30
fix: PR automation bindings

zreigz:
Fix PR automation bindings

## Checklist


go/controller/internal/controller/prautomation_controller.go
```diff
@@ -173,6 +173,10 @@ func (in *PrAutomationReconciler) sync(ctx context.Context, prAutomation *v1alph
 	if err := in.ensure(prAutomation); err != nil {
```
michaeljguarino, line 173: Shouldn't we just move this to above line 155?  We technically call `in.attributes(...)` twice

zreigz: should be fine, done

zreigz:
also have fixed errors after client changes

michaeljguarino (approved):

PR #1254, opened 2024-08-08
add auth token encryption

jsladerman:
encrypts the auth token in local storage


michaeljguarino (commented):
This should include the github actions/docker change to add the env var to the img build

michaeljguarino:
The ideal would ultimately be to have this wired in via a script as well that modifies the built bundle at boot, but we can table that for now

michaeljguarino (approved):

michaeljguarino (approved):

PR #1256, opened 2024-08-08
feat(compatibilities): add containerd

walkoss:
Added [containerd](https://www.cncf.io/projects/containerd/) CNCF graduated project 

Using https://containerd.io/releases/ for compatibility matrix
Versions < 1.6 are deprecated
## Checklist


michaeljguarino (commented):
this is usually installed at a lower level than in the k8s api directly, so I don't think we'd need to include it in the compatibility tables

PR #1257, opened 2024-08-08
feat(compatibilities): add cri-o

walkoss:
Added [cri-o](https://www.cncf.io/projects/cri-o/) CNCF graduated project 

Using https://github.com/cri-o/cri-o?tab=readme-ov-file#compatibility-matrix-cri-o--kubernetes for compatibility matrix
There is a direct compatibility between cri-o and kubernetes versions 1.29 CRIO <> 1.29 k8s and so on 
## Checklist


michaeljguarino (commented):
Since the cri is installed on the node and not in the k8s api directly, i don't think this is going to be usable w/in the compat matrices product

PR #1260, opened 2024-08-08
feat(compatibilities): add karmada

walkoss:
Add [Karmada](https://www.cncf.io/projects/karmada/) CNCF incubating project
Compatibility matrix -> https://github.com/karmada-io/karmada?tab=readme-ov-file#kubernetes-compatibility

## Checklist

### DIFF
--- a/static/compatibilities/karmada.yaml
+++ b/static/compatibilities/karmada.yaml
@@ -1,0 +1,25 @@
+icon: https://avatars.githubusercontent.com/u/74165798?s=2008&v=4
+git_url: https://github.com/karmada-io/karmada
+release_url: https://github.com/karmada-io/karmada/releases/tag/v{vsn}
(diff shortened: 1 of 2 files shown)


static/compatibilities/karmada.yaml
```diff
@@ -0,0 +1,25 @@
+icon: https://avatars.githubusercontent.com/u/74165798?s=2008&v=4
+git_url: https://github.com/karmada-io/karmada
+release_url: https://github.com/karmada-io/karmada/releases/tag/v{vsn}
+helm_repository_url: https://github.com/karmada-io/karmada/tree/master/charts/karmada
```
walkoss, line 4: **question:** will that url work properly on your product @michaeljguarino ?

michaeljguarino:
I believe another user submitted this one, and it's an interesting project but I don't believe it'll be run on all the clusters in the fleet due to its architecture, so it's not an ideal case to support in our compatibility set

PR #1285, opened 2024-08-12
add native helm repo url to services list

jsladerman:
also changes the help icon


assets/src/components/cd/services/ServicesColumns.tsx
```diff
@@ -114,7 +114,11 @@ export const ColRepo = columnHelper.accessor(({ node }) => node, {
           <ColWithIcon
             truncateLeft
             icon={
-              helm ? getProviderIconUrl('byok', theme.mode) : <GitHubLogoIcon />
+              svc?.helm ? (
```
michaeljguarino, line 117: this is slightly wrong, we should really check if there's a helm url at all and show the helm-ish icon if so (you lose that info a bit by having one url var)

jsladerman: I changed it to just check if it has a git repo instead, I assume it's one or the other right?

michaeljguarino: Helm services can have git repos, so really the logic i gave is the best option

jsladerman: ah makes sense. fixed

michaeljguarino (approved):

PR #1318, opened 2024-08-21
feat(controller): add support for stack variables to the InfrastructureStack

floreks:
- Add support for `variables` arg to the `InfrastructureStack`
- Fix an issue with git repository being stuck in deletion when ID was not synchronized before deletion was triggered
- Fix an issue with pr automation sync returning both `err` and `requeue` causing core kubernetes reconciler to throw an error saying that both can't be set at the same time

## Test Plan

`console.plrl-dev-aws.onplural.sh`

## Checklist


go/controller/internal/controller/gitrepository_controller.go
```diff
@@ -144,18 +143,13 @@ func (r *GitRepositoryReconciler) handleDelete(ctx context.Context, repo *v1alph
 	logger := log.FromContext(ctx)
 	if controllerutil.ContainsFinalizer(repo, RepoFinalizer) {
 		logger.Info("delete git repository")
-		if repo.Status.ID == nil {
```
floreks, line 147: This was causing repository to get stuck in a delete loop if it was deleted before ID was synchronized.

go/controller/internal/controller/prautomation_controller.go
```diff
@@ -86,7 +87,7 @@ func (in *PrAutomationReconciler) Reconcile(ctx context.Context, req reconcile.R
 	if err != nil {
 		logger.Error(err, "unable to create or update prAutomation")
 		utils.MarkFalse(prAutomation.SetCondition, v1alpha1.SynchronizedConditionType, v1alpha1.SynchronizedConditionReasonError, err.Error())
-		return requeue, err
```
floreks, line 89: Returning both non-empty `ctrl.Result` and `err` was causing a parent reconciler error. They both can't be provided at the same time.

michaeljguarino (approved):

PR #1339, opened 2024-08-26
service error text wrap

jsladerman:
new behavior wraps text while still preserving whitespace

### DIFF
--- a/assets/src/components/cd/services/service/ServiceErrors.tsx
+++ b/assets/src/components/cd/services/service/ServiceErrors.tsx
@@ -1,22 +1,24 @@
 import {
-  Code,
+  Card,
   EmptyState,
   Table,
   useSetBreadcrumbs,
 } from '@pluralsh/design-system'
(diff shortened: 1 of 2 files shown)


assets/src/components/cd/services/service/ServiceErrors.tsx
```diff
@@ -38,14 +40,22 @@ export const ColMessage = columnHelper.accessor((row) => row?.message, {
...
   meta: { gridTemplate: '6fr' },
-  cell: ({ getValue }) => (
-    <Code
-      maxHeight="300px"
-      width="100%"
-    >
-      {getValue()}
-    </Code>
-  ),
+  cell: function Cell({ getValue }) {
```
michaeljguarino, line 43: how big is the encapsulating modal?  We should definitely expand the width since these messages do get long

jsladerman: fixed, will expand if needed while staying within viewport

michaeljguarino (approved):

PR #1347, opened 2024-08-29
feat: support PLURAL notification sink in router

zreigz:
## Checklist


go/controller/api/v1alpha1/notificationsink_types.go
```diff
@@ -36,6 +36,9 @@ type NotificationSinkSpec struct {
 	// Configuration for the specific type
 	// +kubebuilder:validation:Optional
 	Configuration SinkConfiguration `json:"configuration"`
+
+	// +kubebuilder:validation:Optional
+	Bindings []Binding `json:"bindings,omitempty"`
```
michaeljguarino, line 41: Can we add a docstring here like:

```
// Bindings to determine users/groups to be notified for PLURAL sync types
```

michaeljguarino (approved):

go/controller/api/v1alpha1/notificationsink_types.go
zreigz: done

PR #1356, opened 2024-09-02
feat(controller): add projectref to prautomation crd

floreks:
- Add project ref support to pr automation
- Fix lint by refactoring vulnerable method

## Checklist


michaeljguarino (approved):
what was the lint issue?

floreks:
It was allowing non-const format strings here: https://github.com/pluralsh/console/pull/1356/files#diff-cff252b6484cfaccfc4a9673d1e9b2a3fed73c8e68a924e401d1bdfc9b4ae57cL245-L252

We were usually just passing errors directly as format strings. That could be used to prepare a malicious error message that could be used to extract sensitive data.

PR #1368, opened 2024-09-10
feat: Add secret sharing experience

maciaszczykm:
## Checklist

### DIFF
--- a/assets/src/components/sharesecret/ShareSecretModal.tsx
+++ b/assets/src/components/sharesecret/ShareSecretModal.tsx
@@ -52,9 +52,6 @@ export default function ShareSecretModal({
   >([])
   const disabled = !name || !secret
   const toastRef = useRef<HTMLElement>()
-  const inputRef = useRef<HTMLInputElement>()
-
-  useEffect(() => inputRef.current?.focus?.(), [])
 
   useEffect(() => {
     if (toastRef.current)
@@ -178,7 +175,6 @@ export default function ShareSecretModal({
           >
             <Input
               disabled={completed}
-              inputProps={{ ref: inputRef }}
               value={name}
               onChange={(e) => setName(e.currentTarget.value)}
             />


assets/src/components/sharesecret/ShareSecretModal.tsx
```diff
@@ -0,0 +1,230 @@
...
+  const [secret, setSecret] = useState('')
+  const [bindings, setBindings] = useState<
+    Pick<PolicyBindingFragment, 'user' | 'group'>[]
+  >([])
+  const disabled = !name || !secret
+  const inputRef = useRef<HTMLInputElement>()
+
+  useEffect(() => inputRef.current?.focus?.(), [])
+
+  // TODO: Reset on each open.
```
jsladerman, line 53: with the new modal, state is reset automatically when it opens or closes (because the modal will mount/unmount its children). just need the \<Modal> component to wrap around all the logic- have a recent example [here](https://github.com/pluralsh/plural/blob/61088262a6b733f7fc487ad602783b84fbacc141/www/src/components/overview/clusters/plural-cloud/ConsoleInstanceOIDC.tsx#L60-L70)

only caveat is that there's not a great way for this to work when actions need to share state with the main content. in that example I just put it all into one component instead of having the actions separate. the new modal should also focus the first input by default (unless it's prevented with the onOpenAutoFocus prop)

don't have to do it this way but thought it was worth pointing out

maciaszczykm: Thanks for the reminder. I removed the focus handler because as you said it works out of the box. For the reset I kept current logic as otherwise I would need do some more refactoring (action buttons are specified on modal element).

jsladerman: Makes sense

assets/src/components/sharesecret/ShareSecretModal.tsx
```diff
@@ -0,0 +1,237 @@
...
+  const [toast, setToast] = useState(false)
+  const [name, setName] = useState('')
+  const [secret, setSecret] = useState('')
+  const [bindings, setBindings] = useState<
+    Pick<PolicyBindingFragment, 'user' | 'group'>[]
+  >([])
+  const disabled = !name || !secret
+  const toastRef = useRef<HTMLElement>()
+
+  useEffect(() => {
```
jsladerman, line 56: we should fix this in the DS at some point, generally want to avoid effects like this

jsladerman (approved):

assets/src/components/sharesecret/ShareSecretModal.tsx
maciaszczykm: Totally agree.

PR #1386, opened 2024-09-16
reduce unnecessary log outs

jsladerman:
this was likely causing unnecessary logouts, and we don't necessarily need a logout when LoginError is rendered

### DIFF
--- a/assets/src/components/login/Login.tsx
+++ b/assets/src/components/login/Login.tsx
@@ -1,7 +1,7 @@
-import { useMutation, useQuery } from '@apollo/client'
+import { ApolloError, useMutation, useQuery } from '@apollo/client'
 import { Button, LoopingLogo } from '@pluralsh/design-system'
 import { WelcomeHeader } from 'components/utils/WelcomeHeader'
-import { useMeQuery } from 'generated/graphql'
+import { User, useMeQuery } from 'generated/graphql'
 import gql from 'graphql-tag'
 import { Div, Flex, Form, P } from 'honorable'
 import { RefObject, useEffect, useRef, useState } from 'react'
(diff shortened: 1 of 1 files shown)


assets/src/components/login/Login.tsx
```diff
@@ -42,16 +37,6 @@ const setInputFocus = (ref: RefObject<any>) => {
 }
 
 function LoginError({ error }) {
```
michaeljguarino, line 39: Where is this used again?

michaeljguarino (commented):
if you're trying to fix this: https://github.com/pluralsh/console/blob/3795335073319ab7f1584f9e4a53996b5d57f0be/assets/src/components/login/Login.tsx#L111-L113

Then maybe the better approach is to just trigger the timeout effect if the error is not a network error

jsladerman:
> if you're trying to fix this:
> 
> https://github.com/pluralsh/console/blob/3795335073319ab7f1584f9e4a53996b5d57f0be/assets/src/components/login/Login.tsx#L111-L113
> 
> Then maybe the better approach is to just trigger the timeout effect if the error is not a network error

good point, I updated it to check for the error I was getting when simulating it. seemed to work correctly when testing. figured checking for any network error might be too big of an umbrella

assets/src/components/login/Login.tsx
```diff
@@ -41,16 +41,18 @@ const setInputFocus = (ref: RefObject<any>) => {
...
 
-function LoginError({ error }) {
+function LoginError({ error }: { error: ApolloError | undefined }) {
   useEffect(() => {
-    const to = setTimeout(() => {
-      wipeToken()
-      wipeRefreshToken()
-      window.location = '/login' as any as Location
-    }, 2000)
+    if (error?.networkError?.message !== 'Failed to fetch') {
```
michaeljguarino, line 46: i think it's ok to ignore if it's a network error entirely although do 500's get parsed as network errors?

There's actually a couple discrete errors we send when it doesn't auth:

* if there is no token, you'll get an `unauthenticated` error message
* if it's an expired/invalid jwt, you'll probably get a `401` with an `unauthorized` response.

assets/src/components/login/Login.tsx
```diff
@@ -103,7 +105,6 @@ export function GrantAccess() {
 export function EnsureLogin({ children }) {
   const { data, error, loading } = useMeQuery({
     pollInterval: POLL_INTERVAL,
-    errorPolicy: 'ignore',
```
michaeljguarino, line 106: keep this for sure

assets/src/components/login/Login.tsx
```diff
@@ -41,16 +41,24 @@ const setInputFocus = (ref: RefObject<any>) => {
...
+  me: Nullable<User>
+  error: ApolloError | undefined
+}) {
   useEffect(() => {
-    const to = setTimeout(() => {
-      wipeToken()
-      wipeRefreshToken()
-      window.location = '/login' as any as Location
-    }, 2000)
+    if (!error?.networkError || !me) {
```
michaeljguarino, line 52: I think this should be an `&&` not an `||`

michaeljguarino (approved):

PR #1398, opened 2024-09-18
fix form toggles

jsladerman:
smtp and pr automation forms were both having issues capturing toggle values (smtp form wasn't actually setting the default to "false" so it would register undefined if there was no toggle, and conversely pr automation form was treating false as undefined)

### DIFF
--- a/assets/src/components/pr/automations/prConfigurationUtils.tsx
+++ b/assets/src/components/pr/automations/prConfigurationUtils.tsx
@@ -49,16 +49,17 @@ export function validateAndFilterConfig(
 
     if (conditionMet && name) {
       let value: string | boolean | undefined = configVals[name]
+      const isBool = configItem.type === ConfigurationType.Bool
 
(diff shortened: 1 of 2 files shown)


michaeljguarino (approved):

assets/src/components/pr/automations/prConfigurationUtils.tsx
```diff
@@ -49,16 +49,17 @@ export function validateAndFilterConfig(
...
+      const isBool = configItem.type === ConfigurationType.Bool
 
-      if (configItem.type === ConfigurationType.Bool) {
+      if (isBool) {
         value = parseToBool(value)
       }
 
-      if (value) {
-        filteredValues.push({ name, value })
+      if (value || isBool) {
```
michaeljguarino, line 58: if the default is false, couldn't this also have been solved with something like `value !== undefined`?

PR #1413, opened 2024-09-25
fix: General pagination fixes and improvements

maciaszczykm:
## Checklist


michaeljguarino (commented):
How well did we regression test this?

maciaszczykm:
@michaeljguarino I will do another round of testing as it is quite big but I already checked most important views.

maciaszczykm:
I think I checked all list views now. Fixed issue with selector and other smaller things. I'll merge it once checks will pass.

PR #1417, opened 2024-09-26
Log prometheus queries

michaeljguarino:
Might help w/ prometheus troubleshooting

## Test Plan
n/a

## Checklist


floreks (approved):

lib/prometheus/client.ex
```diff
@@ -17,6 +18,7 @@ defmodule Prometheus.Client do
 
   def query(client \\ nil, query, start, end_t, step, variables) do
     query = variable_subst(query, variables)
+    Logger.info "Issuing prometheus query: #{query}"
```
floreks, line 21: We should probably remove this log not to generate obsolete log lines.

PR #1449, opened 2024-10-09
Catalog CRD support

zreigz:
Catalog CRD support

## Checklist


michaeljguarino:
We should be sure to add create bindings too w/ the schema change done here: https://github.com/pluralsh/console/pull/1458

go/controller/api/v1alpha1/catalog_types.go
```diff
@@ -0,0 +1,104 @@
...
+		attrs.Tags = make([]*console.TagAttributes, 0)
+		for k, v := range c.Spec.Tags {
+			attrs.Tags = append(attrs.Tags, &console.TagAttributes{
+				Name:  k,
+				Value: v,
+			})
+		}
+	}
+
+	if c.Spec.Bindings != nil {
```
michaeljguarino, line 89: Catalogs should also have create bindings too

go/controller/api/v1alpha1/catalog_types.go
```diff
@@ -0,0 +1,104 @@
...
+		attrs.ReadBindings = PolicyBindings(c.Spec.Bindings.Read)
+		attrs.WriteBindings = PolicyBindings(c.Spec.Bindings.Write)
```
michaeljguarino, line 91: will this re-enqueue if the bindings aren't resolved?

zreigz: yes, in the controller:
```
	// Sync Catalog CRD with the Console API
	if err := r.ensure(catalog); err != nil {
		if goerrors.Is(err, operrors.ErrRetriable) {
			utils.MarkCondition(catalog.SetCondition, v1alpha1.SynchronizedConditionType, v1.ConditionFalse, v1alpha1.SynchronizedConditionReasonError, err.Error())
			return requeue, nil
		}
		return ctrl.Result{}, err
	}
```

go/controller/api/v1alpha1/catalog_types.go
zreigz: The create binding is missing in the newest schema.graphql

michaeljguarino: it should be present here: https://github.com/pluralsh/console/pull/1458

go/controller/api/v1alpha1/catalog_types.go
michaeljguarino: Oh i see cool

go/controller/api/v1alpha1/catalog_types.go
zreigz: done

PR #1454, opened 2024-10-09
add email settings tab to profile

jsladerman:
currently just lets you toggle digest emails on and off


assets/src/components/profile/EmailSettings.tsx
```diff
@@ -0,0 +1,66 @@
...
+        >
+          <Switch
+            checked={subscribed}
+            onChange={() => setSubscribed(!subscribed)}
+            css={{
+              flexDirection: 'row-reverse',
+              justifyContent: 'space-between',
+            }}
+          >
+            Subscribe to digest emails
```
michaeljguarino, line 46: this should default to true fwiw (if undefined/null)

jsladerman: updated

michaeljguarino (approved):

PR #1459, opened 2024-10-10
fix secret sharing bugs

jsladerman:
makes it so the user has to explicitly click "consume secret" instead of consuming on render, which seemed to be causing some issues

also fixes a bug that was causing new secrets to generate every time the URL copy button was clicked

Screenshot 2024-10-10 at 1 08 37 PM
Screenshot 2024-10-10 at 1 08 53 PM
Screenshot 2024-10-10 at 1 14 07 PM


michaeljguarino:
Could you post some screengrabs of what it looks like?

jsladerman:
> Could you post some screengrabs of what it looks like?

added

michaeljguarino (approved):

PR #1460, opened 2024-10-11
feat: Show client info for upgrade insights 

maciaszczykm:
Zrzut ekranu 2024-10-11 o 14 13 31


michaeljguarino (approved):
cool, we can go ahead and ship this, but probably should have nick look at it when he's available.

maciaszczykm:
@michaeljguarino I agree, @nickcernera feedback would help to improve it. There is quite a lot of data to show for an expansion panel. Perhaps at some point we should move it to modal or another view.

PR #1470, opened 2024-10-14
feat: Add stack variables tab

maciaszczykm:
Zrzut ekranu 2024-10-14 o 11 40 04


assets/src/components/stacks/variables/StackVariables.tsx
```diff
@@ -0,0 +1,103 @@
...
+  }),
+  columnHelper.accessor((row) => row.value, {
+    id: 'value',
+    header: 'Value',
+    meta: { truncate: true },
+    cell: ({ getValue }) => {
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      const theme = useTheme()
+
+      return <span css={{ ...theme.partials.text.code }}>{getValue()}</span>
```
michaeljguarino, line 42: This can technically be any type, including an object, wondering if we should json stringify it?

maciaszczykm: Right, I've done it.

michaeljguarino (approved):

PR #1473, opened 2024-10-15
add state filter to service components

jsladerman:
Screenshot 2024-10-15 at 6 28 46 PM


michaeljguarino (commented):
Can we not put the chip styling in the selector label itself?

michaeljguarino (approved):

michaeljguarino:
should verify the docker image can actually be built though

PR #1475, opened 2024-10-16
feat: Add raw view to certificate on ingress page

maciaszczykm:


michaeljguarino (approved):
It actually would be pretty nice to add a delete button there at some point to redrive the cert generation process, not sure we have the gql in place to do that yet though.  Can add to backlog.

michaeljguarino:
I think it would be nice to have this be a tad more discoverable than nesting it in an action menu but it works for now.

PR #1495, opened 2024-10-23
feat: introduce ollama

zreigz:
## Checklist

### DIFF
--- a/charts/ollama/Chart.yaml
+++ b/charts/ollama/Chart.yaml
@@ -1,0 +1,21 @@
+apiVersion: v2
+name: ollama
+description: Get up and running with large language models locally.
+
+type: application
+
+version: 0.1.0
+
+appVersion: "0.3.14"
+
+annotations:
+  artifacthub.io/category: ai-machine-learning
+
+kubeVersion: "^1.16.0-0"
+home: https://ollama.ai/
+icon: https://ollama.ai/public/ollama.png
+keywords:
+  - ai
+  - llm
+  - llama
+  - mistral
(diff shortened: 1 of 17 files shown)


charts/ollama/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v2
+name: ollama
```
michaeljguarino, line 2: I'm guessing this was just derived by `helm pull` ing the ollama chart right?  Could we add a make target for that so it's accessible in the future and perhaps we could even have a github actions cron run to do it and pull request if any changes are found.

michaeljguarino (commented):
is the `publish-ollama.yaml` action supposed to be empty?

charts/ollama/Chart.yaml
zreigz: done

.github/workflows/publish-ollama.yaml
```diff
@@ -0,0 +1,45 @@
...
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Login to GHCR
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
```
michaeljguarino, line 37: I think this action should ultimately be a matrix gh action with the model variants we want to support.  We can start small w/ just llama3.1 8b i think and then add additional images for other ones.

michaeljguarino (approved):

PR #1496, opened 2024-10-23
enable console metrics

seemywingz:
enable pod metrics
## Checklist


michaeljguarino (commented):
we don't want to enable this by default

michaeljguarino (commented):
this shouldn't be done by default in the chart, it will make it uninstallable on clusters not including the prometheus operator.  I've already enabled this in the infra monorepo.

PR #1506, opened 2024-10-24
wip: feat: add ai proxy

floreks:
```
TBD
```

## Test Plan

- [ ] Locally
- [ ] Test on dev

## Checklist


michaeljguarino (commented):
we should write at least one basic mock test against the endpoint to ensure regressions are covered and make sure there's a gh action to run it

go/ai-proxy/api/openai/openai.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+func toFlatMessage(choices []openai.Choice) ollamaapi.Message {
+	var result ollamaapi.Message
+	if len(choices) == 0 {
+		return result
+	}
+
+	choice := choices[0]
```
michaeljguarino, line 85: yeah this is basically the same logic I'm using in the main console so should work.

floreks:
> we should write at least one basic mock test against the endpoint to ensure regressions are covered and make sure there's a gh action to run it

Yeah, I also wanted to add some basic tests and gh actions. Should be mostly ready tomorrow.

charts/ai-proxy/values.yaml
```diff
@@ -0,0 +1,113 @@
...
+image:
+  repository: ghcr.io/pluralsh/ai-proxy
+  # This sets the pull policy for images.
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "v1.0.0"
+
+config:
+  provider: openai
+  providerHost: changeme
```
michaeljguarino, line 18: should this be a correct default?

floreks: I guess I can use openai api as the default

PR #1507, opened 2024-10-24
hide insights tab on components that have none

jsladerman:
will also fix the landing page issue


michaeljguarino (approved):
will this also make it not the default tab?

jsladerman:
> will this also make it not the default tab?

yeah, that was only happening because it goes to the first tab in the directory (which on unstructured components would've been insights since they had no info). I put the insights tab after raw which should always be there

PR #1518, opened 2024-10-28
feat: Add indicators to dashboard filters

maciaszczykm:
## Checklist

### DIFF
--- a/assets/src/components/kubernetes/Navigation.tsx
+++ b/assets/src/components/kubernetes/Navigation.tsx
@@ -114,7 +114,7 @@ export default function Navigation() {
           flexShrink: 1,
           height: '100%',
           width: '100%',
-          overflow: 'hidden',
+          minWidth: 0,
         }}
       >
         <div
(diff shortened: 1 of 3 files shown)


assets/src/components/kubernetes/Navigation.tsx
```diff
@@ -114,7 +114,7 @@ export default function Navigation() {
           flexShrink: 1,
           height: '100%',
           width: '100%',
-          overflow: 'hidden',
+          minWidth: 0,
```
maciaszczykm, line 117: @jsladerman This solves issue with overflows getting clipped. `overflow: hidden` was used there to ensure that tab scroll will work correctly. Instead we can set `minWidth: 0` as by default flex provides its own value.

assets/src/components/kubernetes/Navigation.tsx
```diff
@@ -114,7 +114,7 @@ export default function Navigation() {
...
         }}
```
maciaszczykm, line 118: @jsladerman This solves issue with overflows getting clipped. `overflow: hidden` was used there to ensure that tab scroll will work correctly but we can set `minWidth: 0` instead (`display: flex` sets automatic min size).

jsladerman (approved):

assets/src/components/kubernetes/Navigation.tsx
jsladerman: Oh interesting- nice job

PR #1521, opened 2024-10-28
get latest kube version and update KUBE_VERSION file before running 

seemywingz:
get latest kube version and update KUBE_VERSION file before running scrapers

## Checklist


utils/compatibility/utils.py
```diff
@@ -62,7 +62,14 @@ def latest_kube_version():
         latest = response.text.lstrip("v")
         latest = latest.split(".")
```
michaeljguarino, line 63: this should probably use a python semver parser and ignore rc versions too.

michaeljguarino (approved):

michaeljguarino:
actually looks like a test needs to be fixed too

utils/compatibility/utils.py
seemywingz: ack, merging these changes into the other PR to use the semver util:
https://github.com/pluralsh/console/pull/1524

PR #1524, opened 2024-10-28
add compatibility scraper for karpenter and util `validate_semver`

seemywingz:
add compatibility scraper for karpenter and util `validate_semver`

## Checklist


utils/compatibility/utils.py
```diff
@@ -52,6 +72,21 @@ def write_yaml(file_path, data):
...
 
 
+def validate_semver(version_str):
+    try:
+        # Coerce the version string to handle versions like "1.30", "1.27", or "1.2.3"
+        version = semantic_version.Version.coerce(version_str)
+
+        # Check if prerelease and build are empty
+        if version.prerelease or version.build:
+            return False
```
michaeljguarino, line 82: `return None` would probably be a slightly more natural typing for this function (shouldn't affect the behavior any though)

michaeljguarino (commented):
mostly lgtm, minor fix though

utils/compatibility/utils.py
seemywingz: Ack, that makes sense.

michaeljguarino:
The test that's failing specifies a specific kube version, and just needs to be rewritten now that it's moved to 1.31.  Make that change and this is good to merge.

test/console/graphql/queries/deployments/cluster_queries_test.exs
```diff
@@ -583,6 +583,7 @@ defmodule Console.GraphQl.Deployments.ClusterQueriesTest do
       admin = admin_user()
       insert_list(2, :cluster, current_version: "1.29", pinged_at: Timex.now(), upgrade_plan: %{compatibilities: true, incompatibilities: true, deprecations: true})
       insert_list(3, :cluster, current_version: "1.30", pinged_at: Timex.now() |> Timex.shift(days: -1))
+      insert_list(2, :cluster, current_version: "1.31", pinged_at: Timex.now() |> Timex.shift(days: -1))
```
michaeljguarino, line 586: the reason you had to change the asserts is you didn't modify the top set of clusters to `1.30` and the following to `1.31` to preserve the version pattern

seemywingz: this looks better: https://github.com/pluralsh/console/pull/1524/files#diff-9356619c481fa6b7722ae779fe4a8d30d3038acfe35db254f7db13039f6e0cee

michaeljguarino (approved):

PR #1546, opened 2024-11-04
enhancement: pass raw kubernetes object to explain w/ ai prompt

maciaszczykm:
In Kubernetes Dashboard API raw objects are not part of resource details. To fetch raw objects we are using separate query. Because of that we can reduce number of `useExplainWithAI` calls to just two. One for CD components and one for all Kubernetes resource details.

### DIFF
--- a/assets/src/components/kubernetes/common/ResourceDetails.tsx
+++ b/assets/src/components/kubernetes/common/ResourceDetails.tsx
@@ -1,7 +1,8 @@
 import { ReactElement, ReactNode, useMemo, useRef, useState } from 'react'
 import { SubTab, TabList } from '@pluralsh/design-system'
 import { useTheme } from 'styled-components'
-import { useMatch, useResolvedPath } from 'react-router-dom'
+import { useMatch, useParams, useResolvedPath } from 'react-router-dom'
+import pluralize from 'pluralize'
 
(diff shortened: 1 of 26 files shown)


assets/src/components/kubernetes/common/ResourceDetails.tsx
```diff
@@ -36,6 +46,33 @@ export default function ResourceDetails({
...
+    client: KubernetesClient(clusterId ?? ''),
+    skip: !clusterId || kind === 'secret',
+    fetchPolicy: 'no-cache',
+    variables: { kind, name, namespace } as ResourceQueryVariables &
+      NamespacedResourceQueryVariables,
+  })
+
+  const prompt = useMemo(() => {
+    return data?.handleGetResource?.Object
+      ? `Describe the following Kubernetes ${kind} resource: ${JSON.stringify(data?.handleGetResource?.Object)}`
```
michaeljguarino, line 71: Just double-checking, the capitalization is correct here?

maciaszczykm: It's not actually. Do we need it?

michaeljguarino: oh i mean is there a field called `Object` instead of `object`?

PR #1548, opened 2024-11-04
feat: add base url to openai settings

maciaszczykm:
## Checklist

### DIFF
--- a/assets/src/components/settings/global/GlobalSettingsAIProviders.tsx
+++ b/assets/src/components/settings/global/GlobalSettingsAIProviders.tsx
@@ -51,6 +51,7 @@ export function initialSettingsAttributes(
           ? {
               openai: {
                 model: ai.openai.model,
+
                 accessToken: '',
               },
             }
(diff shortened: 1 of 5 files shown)


assets/src/components/settings/global/GlobalSettingsAIProviders.tsx
```diff
@@ -94,19 +95,74 @@ export function validateAttributes(
...
+      >
+        <Input
+          disabled={!enabled}
+          value={settings?.model}
+          onChange={(e) => {
+            updateSettings({ model: e.currentTarget.value })
+          }}
+        />
+      </FormField>
+      <FormField
```
michaeljguarino, line 126: This field should be optional, will the form still enable saving if blank?

maciaszczykm: @michaeljguarino Yes it does.

maciaszczykm: Only access token is required for OpenAI at the moment.

PR #1566, opened 2024-11-06
update add service account file drop to vertex ai settings

jsladerman:
also updates DS


assets/src/components/settings/global/GlobalSettingsAIProviders.tsx
```diff
@@ -99,7 +103,7 @@ export function validateAttributes(
         settings.bedrock?.secretAccessKey
       )
     case AiProvider.Vertex:
-      return true
+      return !!settings.vertex?.model
```
michaeljguarino, line 106: if you rebase on my current pr, this shouldn't be required anymore

michaeljguarino: you can just pr against my branch as well so they get merged together

michaeljguarino (approved):

PR #1616, opened 2024-11-25
change Ready condition on success

zreigz:
Check InfrastructureStack status to set the real Ready condition

## Checklist


go/controller/internal/client/stack.go
```diff
@@ -27,6 +27,36 @@ func (c *client) GetStack(ctx context.Context, id string) (*console.Infrastructu
 	return response.InfrastructureStack, err
 }
 
+func (c *client) GetStackId(ctx context.Context, id string) (*console.InfrastructureStackIDFragment, error) {
```
michaeljguarino, line 30: rename this to `GetStackById` I think

zreigz: done

michaeljguarino (approved):

PR #1625, opened 2024-11-27
bump flux to 2.14.0

zreigz:
bump flux to 2.14.0

## Checklist


charts/console/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: console
 description: >-
   deploys the plural console and additional dependencies, for use in bring-your-own-kube setups
 appVersion: 0.10.40
-version: 0.3.68
+version: 0.3.69
```
michaeljguarino, line 6: don't bump this please, needs to be bumped in the release

michaeljguarino (changes requested):

charts/console/Chart.yaml
zreigz: done

michaeljguarino (approved):

PR #1626, opened 2024-11-27
improve socket connect retry logic

jsladerman:
console wasn't correctly detecting when the phoenix socket dropped


assets/src/helpers/client.ts
```diff
@@ -60,14 +60,12 @@ export function buildClient(gqlUrl, wsUrl, fetchToken) {
   })
 
   setInterval(() => {
-    if (socket.transport.readyState == WebSocket.CLOSED) {
+    if (!socket.conn) {
```
michaeljguarino, line 63: are you sure this ever goes undefined?

jsladerman: it happened every time i called socket.disconnect() when testing

michaeljguarino (approved):

PR #1664, opened 2024-12-09
update reduce_versions to only add latest, new major or new minor if …

seemywingz:
…k8s compat chamged

## Test Plan

I ran the scaper and manually verified the version reduction is what we are looking for

## Checklist


seemywingz:
these are the main changes: https://github.com/pluralsh/console/compare/PROD-2899?expand=1#diff-8edc21f1bef86cfd5da90125c21acdc7ea014bcde87378597911a37d37a326e4

michaeljguarino (approved):

PR #1673, opened 2024-12-12
feat: add GeneratedSecret controller

zreigz:
- Check if secrets exist (from the destination paths), if not create a new one.
 - Template secrete values

## Checklist


zreigz:
I have to extend a schema to store the result in the status block to be able to make an update

michaeljguarino (commented):
Couple of things:

* We should store the rendered template in a secret mapped into the generatedsecret status field.  That way we can rereconcile drifted secrets in the future.
* we should have a way to use a Watches on the destination secret (label selector?) that way we can detect if the data field has been modified differently from the templated value, and then remediate that.

michaeljguarino (approved):

PR #1688, opened 2024-12-16
update aws efs scraper

seemywingz:
## Test Plan

manually tested the changes locally and verified the scraper has the desired output

## Checklist


michaeljguarino (approved):
This is probably fine, but i feel like there might be a way to just have this take the last known efs driver version as the first in the list always.  Snce the intermediate versions aren't actually that helpful.

PR #1752, opened 2025-01-14
Add functionality to AI Proxy for full Open AI protocol

JohnBlackwell:
Supports full implementation of the `openai chat completions api` including tool calling and SSE.


michaeljguarino (commented):
We should make a separate provider type for the OpenAI endpoint, and ideally it should be general enough to support a bedrock backend.

go/ai-proxy/proxy/openai/openai.go
```diff
@@ -0,0 +1,102 @@
...
+			r.Out.URL.Path = targetURL.Path
+
+			klog.V(log.LogLevelDebug).InfoS(
+				"proxying request",
+				"from", fmt.Sprintf("%s %s", r.In.Method, r.In.URL.Path),
+				"to", r.Out.URL.String(),
+			)
+		},
+
+		ModifyResponse: func(resp *http.Response) error {
```
michaeljguarino, line 64: i suspect this is unnecessary and honestly we'd want to keep the response gzipped if possible

go/ai-proxy/test/helpers/common.go
```diff
@@ -10,20 +10,40 @@ import (
...
-	p, err := proxy.NewTranslationProxy(args.Provider(), args.ProviderHost(), args.ProviderCredentials())
-	if err != nil {
-		return nil, err
+	provider := args.Provider()
+	host := args.ProviderHost()
+	creds := args.ProviderCredentials()
+
+	router := mux.NewRouter()
+
+	if provider == api.ProviderOpenAIStandard {
```
michaeljguarino, line 29: there should not be a new openai provider here.  We need to have both router functions running in parallel since it needs to support the ollama and openai endpoitns simultaneously (some cloud console's will still be configured against ollama).

go/ai-proxy/proxy/openai/openai.go
JohnBlackwell: Yeah you're right - it's an optional function and this works fine without it.

go/ai-proxy/proxy/openai/openai.go
```diff
@@ -0,0 +1,62 @@
...
+	token string
+}
+
+func (o *OpenAIProxy) Proxy() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		o.proxy.ServeHTTP(w, r)
+	}
+}
+
+func NewOpenAIStandardProxy(host, token string) (api.OpenAIProxy, error) {
```
michaeljguarino, line 27: let's not use this OpenAIStandard naming anymore

go/ai-proxy/api/openai_standard/openai_standard.go
```diff
@@ -0,0 +1,7 @@
+package openai_standard
```
michaeljguarino, line 1: don't use openai standard naming here either, can probably just use the old openai package and put the const there

michaeljguarino (approved):

PR #1758, opened 2025-01-16
feat: add bootstraptoken controller

floreks:
Added a new `BootstrapToken` controller/crd. It can generate a secret with the bootstrap token that can optionally be tied to the user.

## Test Plan

Locally

## Checklist


michaeljguarino (approved):

go/controller/config/rbac/role.yaml
```diff
@@ -130,3 +133,9 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - deployments.plural.sh
+  resources:
+  - secrets
```
michaeljguarino, line 139: fwiw this looks wrong?

floreks: Ye, noticed that in the controller kube builder comment. It is fixed now.

PR #1773, opened 2025-01-21
Bootstrap tokens should be able to query deploy tokens

michaeljguarino:
This is blocking edge setup currently

## Test Plan
unit tests

## Checklist


maciaszczykm (approved):
Thank you. Will cluster query work as well? It may be needed to redeploy agent in some cases.

michaeljguarino:
> Thank you. Will cluster query work as well? It may be needed to redeploy agent in some cases.

It might not, i can add that

PR #1795, opened 2025-01-27
fix: use updated table component

maciaszczykm:
Use updated table component from https://github.com/pluralsh/design-system/pull/676.


jsladerman (approved):

michaeljguarino (approved):
Are all the tables actually using that CSS or is there a stray goofy one?

michaeljguarino:
Need to make sure dkr builds first though

maciaszczykm:
@michaeljguarino I have fixed issues with types that were failing in Docker build. I think that CSS part was required at some point but now it looks to be redundant.

PR #1827, opened 2025-01-30
Elasticsearch docker test

kinjalh:
Add an unit test that uses a local elasticsearch index to check if it is able to retrieve logs from.
This requires spinning up an elasticsearch container, creating an index on it, adding log entries to the index, querying the index for the logs, and then deleting the index.

## Test Plan
`mix test test/console/graphql/queries/observability_queries_test.exs`

## Checklist


test/console/graphql/queries/observability_queries_test.exs
```diff
@@ -192,6 +196,54 @@ defmodule Console.GraphQl.ObservabilityQueriesTest do
...
+      # We want a fresh index so delete the index if it already exists
+      # Just in case test fails we want to make sure next run doesn't reuse old index
+      with true <- ES.index_exists?(@elastic_cluster_url, @elastic_index_name) do
+        ES.delete_index(@elastic_cluster_url, @elastic_index_name)
+      end
+
+      ES.create_index(@elastic_cluster_url, @elastic_index_name)
+
+      ES.index_documents(@elastic_cluster_url, @elastic_index_name, [
+        %{
```
michaeljguarino, line 212: let's move these documents to helper function in elastic utils, eg something like:

```elixir
def log_document(%Service{}, line) do
  ...map data for log
end
```

And that way you can just index by doing something like:

```elixir
log_document(svc, "valid log") |> index()
log_document(svc, "valid log2") |> index()
log_document(svc2, "valid log") |> index()
refresh()
```

michaeljguarino: you should also just import the elastic helper so it's quicker to call the functions I think

docker-compose.yml
```diff
@@ -8,6 +8,16 @@ services:
       POSTGRES_PASSWORD: postgres
     volumes:
     - database_data:/var/lib/postgresql/data
+  es:
```
michaeljguarino, line 11: i think this might need a volume to really behave nicely

lib/mix/tasks/elasticsearch_setup.ex
```diff
@@ -0,0 +1,32 @@
+# lib/mix/tasks/elasticsearch_setup.ex
```
michaeljguarino, line 1: rm

test/console/graphql/queries/observability_queries_test.exs
```diff
@@ -192,6 +196,33 @@ defmodule Console.GraphQl.ObservabilityQueriesTest do
...
     end
 
+    test "it can fetch logs from an elasticsearch index" do
+      user = insert(:user)
+      svc = insert(:service, read_bindings: [%{user_id: user.id}])
+
+      # the index gets set up using a mix task before the test is run, so we can index directly
+      log_document(svc, "valid log message") |> index_doc(@elastic_cluster_url, @elastic_index_name)
+      log_document(svc, "another valid log message") |> index_doc(@elastic_cluster_url, @elastic_index_name)
+      refresh(@elastic_cluster_url, @elastic_index_name)
```
michaeljguarino, line 206: I'd move the cluster url into your elasticsearch helper module as a default arg (so users don't need to specify it w/ every call to `index_doc` or `refresh`

michaeljguarino: use default args in `refresh` here too

michaeljguarino (approved):

PR #1831, opened 2025-01-31
fix: Deselect namespace on cluster change

maciaszczykm:
## Checklist


michaeljguarino (commented):
what i also saw was the namespace list was the same on cluster change

michaeljguarino (approved):

PR #1842, opened 2025-02-03
feat: add failing service

floreks:
```
TBD
```

## Checklist


go/go.work
```diff
@@ -1,11 +1,10 @@
...
-toolchain go1.23.2
+go 1.23.4
 
 use (
 	./ai-proxy
 	./client // github.com/pluralsh/console/go/client
 	./controller // github.com/pluralsh/console/go/controller
 	./oci-auth // github.com/pluralsh/console/go/oci-auth
 	./tools // github.com/pluralsh/console/go/tools
+	./test/failing-service // github.com/pluralsh/console/go/test/failing-service
```
michaeljguarino, line 9: we should make the subfolder here be `hack` and not test (which to me means unit tests)

floreks: ye, I was thinking about some other names like `mock` or `demo` too

kinjalh: fwiw i'm putting it under a `demo/failing-service` directory

floreks:
closing in favor of #1862

PR #1854, opened 2025-02-05
Fix pipeline gate rerunning

michaeljguarino:
This logic was actually backwards, causing pipelines to repromo gates continuously until satisfied.  the correct logic is to only reopen them if they haven't yet been opened during the promotion.  Also added a test (which wasn't done originally for this to cut corners).

## Test Plan
unit tests

## Checklist


.github/workflows/k8s-versions-updater.yaml
maciaszczykm, line 1: Should it be removed? Does it open PRs too often?

maciaszczykm (approved):

PR #1858, opened 2025-02-06
improve DAG views

jsladerman:
generally consolidates the logic and improves usability/layouts for our React Flow views, which were basically unusable when they got too big

- makes views fit correctly on initial graph render
- changes reset view behavior to show the whole graph in view, along with allowing user to zoom out further if needed
- simplifies the way graphs control specific options (uses dagre options described here https://github.com/dagrejs/dagre/wiki#configuring-the-layout)


michaeljguarino:
this largely looks pretty good, we can at least iterate on it some more where needed

michaeljguarino (approved):

PR #1862, opened 2025-02-06
Failing service demo

kinjalh:
Create a demo service in go that is intentionally flaky (i.e. sometimes returns ISE instead of 200) for a very basic HTTP endpoint.

Set up prometheus metrics and containerize the API server and metrics server, as well as a sidecar container to simulate network traffic.

## Test Plan
tested using docker-compose and kubectl

## Checklist


test-apps/flaky-service/prometheus.yaml
kinjalh, line 1: https://github.com/prometheus-operator/prometheus-operator/blob/main/example/user-guides/getting-started/example-app-pod-monitor.yaml

test-apps/flaky-service/flaky-service.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: v1
+kind: Pod
```
kinjalh, line 2: don't do a Pod, do a deployment instead

static/k8s-versions/eks.json
```diff
@@ -1,6 +0,0 @@
-1.23
```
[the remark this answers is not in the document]
michaeljguarino: (this is what's causing the console test to fail fwiw)

.github/workflows/demo-cd.yaml
```diff
@@ -40,6 +40,7 @@ jobs:
         with:
           images: |
             ghcr.io/pluralsh/flaky-service
+            ghcr.io/pluralsh/flaky-service-sidecar
```
michaeljguarino, line 43: for the sidecar image, you can just use a busybox image and wire in a command that calls curl in a loop by running `/bin/bash -c "..."`

.github/workflows/demo-cd.yaml
```diff
@@ -66,3 +66,16 @@ jobs:
           build-args: |
             GIT_COMMIT=${{ github.sha }}
             VERSION=${{ steps.meta.outputs.version }}
+      - uses: docker/build-push-action@v5
```
michaeljguarino, line 69: this is going to overwrite your sidecar container w/ the poll loop you wrote fwiw

kinjalh: yeah realized that and added it as a separate job

lib/mix/tasks/elasticsearch_setup.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Mix.Tasks.ElasticsearchSetup do
+  use Mix.Task
+  require Tesla
+
+  @elastic_cluster_url Application.compile_env(:elasticsearch_test, :cluster_url)
```
michaeljguarino, line 5: think these files might be git leftovers (I moved them to a folder under `mix/tasks/elasticsearch`

go/demo/flaky-service/api_caller.sh
```diff
@@ -0,0 +1,50 @@
...
+
+  # If the response is empty, log a message
+  if [ -z "$RESPONSE" ]; then
+    echo "Received empty response from http://${ENDPOINT}"
+  else
+    echo "Response: $RESPONSE"
+  fi
+
+  CALL_NUM=$((${CALL_NUM:-0}+1))
+  if [ $((${CALL_NUM} % 10)) == 0 ]; then
```
michaeljguarino, line 38: you shouldn't need to query the metrics endpoint here really

floreks:
The `Dockerfile.sidecar` and bash script could probably be removed and this logic can be directly embedded into the deployment sidecar container based on busybox as @michaeljguarino suggested.

michaeljguarino (approved):

PR #1927, opened 2025-02-19
enhancement: Global service view enhancements

maciaszczykm:
## Checklist


maciaszczykm:
@nickcernera Good catch, can you check now? I've removed min width from service errors column as well.

nickcernera (approved):

PR #1934, opened 2025-02-20
Support openai embeddings endpoints in ai proxy

JohnBlackwell:
## Checklist


michaeljguarino (commented):
The router here still doesn't really make sense to me.  Does it not direct /v1/embeddings to bedrock if that's the underlying provider?  And same for the chat requests.  We shouldn't be exposing bedrock url paths, it should just be the standard openai url paths.

go/ai-proxy/main.go
```diff
@@ -20,43 +19,34 @@ import (
...
-	if err != nil {
-		klog.ErrorS(err, "Could not create proxy")
-		os.Exit(1)
-	}
-
-	eop, err := proxy.NewOpenAIEmbeddingsProxy(api.ProviderOpenAI, args.ProviderHost(), args.ProviderCredentials())
-	if err != nil {
-		klog.ErrorS(err, "Could not create proxy")
-		os.Exit(1)
+		if args.Provider() == api.ProviderBedrock {
```
michaeljguarino, line 25: `if args.Provider() != api.ProviderBedrock`

go/ai-proxy/test/bedrock/bedrock_test.go
```diff
@@ -153,7 +152,7 @@ func TestBedrockEmbeddingsProxy(t *testing.T) {
 		{
 			Name:     "embeddings request should return correct openai response",
 			Method:   "POST",
-			Endpoint: bedrock.EndpointEmbeddings,
+			Endpoint: "/v1/embeddings",
```
michaeljguarino, line 155: thius needs to be on the same prefix as the other openai endpoints

go/ai-proxy/main.go
```diff
@@ -20,43 +19,34 @@ import (
...
+		}
+	} else {
+		router.HandleFunc(ollama.EndpointChat, p.Proxy())
 	}
 
-	bp, err := proxy.NewBedrockProxy(api.ProviderBedrock, args.ProviderCredentials())
-	if err != nil {
-		klog.ErrorS(err, "Could not create proxy")
-		os.Exit(1)
+	if args.Provider() == api.ProviderOpenAI || args.Provider() == api.ProviderBedrock {
```
michaeljguarino, line 35: encapsulate this boolean into a function or some global config, ideally.

go/ai-proxy/main.go
JohnBlackwell: if it's bedrock i want it NoOp bc there's no Ollama to Bedrock proxy and therefore it'll always fail. But if it's _not_ bedrock and there's an err, it should fail becasue then that's truly unexpected

go/ai-proxy/test/bedrock/bedrock_test.go
JohnBlackwell: it's not? https://github.com/pluralsh/console/pull/1934

go/ai-proxy/main.go
michaeljguarino: yea and if there was just an if not clause you don't need the goofy else

go/ai-proxy/test/bedrock/bedrock_test.go
michaeljguarino: the other endpoints exposed were `/openai/v1/chat/completions`.  You can put it under both, but the openai path prefix is not a bad way to do it.

michaeljguarino (approved):

PR #1941, opened 2025-02-26
new log filter ui

jsladerman:
incorporates new SegmentedInput component (https://github.com/pluralsh/design-system/pull/687) to handle date/time filters 
Screenshot 2025-02-26 at 1 23 54 AM


michaeljguarino (approved):

michaeljguarino:
So I actually find the date input ux here quite confusing, i think it would be a lot clearer if it just showed the full current time by default and made it immediately editable.  if there's a concern about getting the "now" time right, just have it constantly tick until its edited manually.

PR #1973, opened 2025-03-06
Softserve setup

kinjalh:
set up softserve git server (https://github.com/charmbracelet/soft-serve) to host the deployment-operator and scaffolds repos such that if the user desires, they don't have to connect to github.

https://linear.app/pluralsh/issue/PROD-3247/implement-self-hostability-for-core-plural-deployment-operator-and

## Test Plan
In `softserve-cd.yaml`, initially had it set to `push` instead of `pull_request` - passing wf: https://github.com/pluralsh/console/actions/runs/13704813366

image publish on ghcr: https://github.com/pluralsh/console/pkgs/container/softserve

## Checklist

### DIFF
--- a/charts/console/templates/softserve.yaml
+++ b/charts/console/templates/softserve.yaml
@@ -1,0 +1,86 @@
+{{- if .Values.gitServer }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: softserve
+  labels:
+{{ include "console.labels" . | indent 4 }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: softserve
+  template:
+    metadata:
+      name: softserve
+      labels:
+        app.kubernetes.io/name: softserve
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        {{ if .Values.podLabels }}
+        {{ toYaml .Values.podLabels | nindent 8 }}
+        {{ end }}
+      {{ if .Values.podAnnotations }}
+      annotations:
+        {{ toYaml .Values.podAnnotations | nindent 8 }}
+      {{ end }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}     
+      {{ if .Values.console.securityContext }}
+      securityContext:
+      {{ toYaml .Values.console.securityContext | nindent 8 }}
+      {{ end }}
+      {{- with .Values.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+      - name: softserve
+        {{ if .Values.console.containerSecurityContext }}
+        securityContext:
+        {{- toYaml .Values.console.containerSecurityContext | nindent 10 }}
+        {{ end }}
+        image: "ghcr.io/pluralsh/softserve:{{ .Chart.AppVersion }}"
+        imagePullPolicy: Never
+        ports:
+          - containerPort: {{ .Values.gitServer.sshPort }}
+          - containerPort: {{ .Values.gitServer.httpPort }}
+          - containerPort: {{ .Values.gitServer.metricsPort }}
+          - containerPort: {{ .Values.gitServer.gitPort }}
+        resources:
+        {{- toYaml .Values.gitServer.resources | nindent 10 }}
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: softserve
+spec:
+  type: ClusterIP
+  selector:
+    app.kubernetes.io/name: softserve
+  ports:
+    - port: {{ .Values.gitServer.sshPort }}
+      targetPort: {{ .Values.gitServer.sshPort }}
+      name: ssh
+    - port: {{ .Values.gitServer.httpPort }}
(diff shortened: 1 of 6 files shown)


softserve/softserve.yaml
```diff
@@ -0,0 +1,63 @@
...
+      app: softserve
+  template:
+    metadata:
+      name: softserve
+      labels:
+        app: softserve
+    spec:
+      containers:
+        - name: softserve
+          image: ghcr.io/pluralsh/softserve:latest
```
michaeljguarino, line 18: we would want this to have the actual version templated via helm

softserve/softserve.yaml
```diff
@@ -0,0 +1,63 @@
...
+  name: softserve
+spec:
+  type: NodePort
+  selector:
+    app: softserve
+  ports:
+    - protocol: TCP
+      port: 23231
+      targetPort: 23231
+      nodePort: 30031
```
michaeljguarino, line 47: do we need to use a nodeport service?  That's not ideal

michaeljguarino:
Github review was goofy, the kubernetes manifests need to live in `charts/console/templates` (where our helm chart is).  It'll need to be default disabled, and the image tag should wire in with the `{{ .Chart.AppVersion }}` field

softserve/softserve.yaml
kinjalh: was using nodeport to test (using kubectl on my local machine) but can change to clusterIP as customer usecase would have the git server running in same k8s cluster (correct me if i'm wrong here)

softserve/Dockerfile.softserve
```diff
@@ -0,0 +1,50 @@
...
+# golang install
+RUN add-apt-repository -y ppa:longsleep/golang-backports
+RUN apt update -y
+RUN apt install -y golang-go
+
+# create keys
+RUN apt install -y openssh-client
+
+# soft-serve install
+RUN go install github.com/charmbracelet/soft-serve/cmd/soft@latest
```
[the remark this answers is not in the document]
kinjalh: it has an issue with the Admin role. Basically, if i use their dockerfile then it starts the `soft sever` server without allowing me to first set the admin ssh key, meaning i am unable to import repos.

softserve/softserve.yaml
michaeljguarino: that's correct, it'd run side-by-side with the console instance querying it

softserve/softserve.yaml
kinjalh: will do

softserve/Dockerfile.softserve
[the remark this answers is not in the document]
kinjalh: https://github.com/charmbracelet/soft-serve/issues/497 - currently an open issue which has some potential PR fixes but they are not yet merged in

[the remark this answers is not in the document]
michaeljguarino: Ok, then we probably want to use a more stripped down base image than ubuntu since it likely has a ton of CVEs.  Try an alpine base first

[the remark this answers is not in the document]
kinjalh: ok will try that, hopefully packages for apk are up to date

charts/console/values.yaml
```diff
@@ -241,3 +241,6 @@ flux2:
     create: true
   policies:
     create: false
+
+localgitserver:
```
michaeljguarino, line 245: let's just make this `gitServer`

charts/console/templates/softserve.yaml
```diff
@@ -0,0 +1,56 @@
+{{- if .Values.localgitserver }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: softserve
```
michaeljguarino, line 5: you're going to need to add a lot of the additional templating you see in the other Deployment objects here, in particular, adding pod/deployment labels, and adding security contexts, as all of those are common enterprise reconfigurations

charts/console/templates/softserve.yaml
```diff
@@ -0,0 +1,56 @@
...
+      containers:
+        - name: softserve
+          image: ghcr.io/pluralsh/softserve:{{ .Chart.AppVersion }}
+          imagePullPolicy: Never
+          ports:
+            - containerPort: 23231
+            - containerPort: 23232
+            - containerPort: 23233
+            - containerPort: 9418
+          resources:
```
michaeljguarino, line 26: this should be moved to the values file and templated in

charts/console/templates/softserve.yaml
kinjalh: sure. Also just to confirm - the deployment already has labels i believe (unless referring to different label)? For the security context i can see some templating in the deployments.yaml file (in the same directory), guessing that I can go off that as a sort of example?

michaeljguarino: yeah, although it's still missing the following:

* pod annotations need to be able to be set
* the deployment should probably have the default label set on it
* container security context should be set
* would also be good to be able to template nodeSelector, affinity and tolerations, which all control pod placement logic

charts/console/templates/softserve.yaml
michaeljguarino: (I was pointing to the resources sub-block fwiw, although the ports are nice too)

kinjalh: sure i'll do for both

charts/console/templates/softserve.yaml
```diff
@@ -0,0 +1,84 @@
+{{- if .Values.gitServer }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: softserve
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: softserve
```
michaeljguarino, line 10: I actually think this might goof up now that you changed the labels in template.metadata.labels to be different from `app: softserve`.  Also the deployment object itself should have the default lables like in other ones.

PR #1974, opened 2025-03-06
Dynamicllay render service account for bedrock

JohnBlackwell:
## Checklist


charts/ai-proxy/templates/deployment.yaml
```diff
@@ -24,7 +24,7 @@ spec:
         {{- end }}
     spec:
       {{- if .Values.serviceAccount.create }}
-      serviceAccountName: ai-proxy
```
michaeljguarino, line 27: we should be able to just use the fixed name w/o any issue

PR #1988, opened 2025-03-10
Softserve Fix

kinjalh:
softserve dockerfile bigfix - install git before running server

## Test Plan
from the `dockerfiles` directory:
to build and run container
```
docker build -t git-server -f Dockerfile.softserve .`
docker run -d -p 23231:23231 -p 23232:23232 --name git-server git-server
```
then should be able to clone:
git:
```
git clone git clone http://localhost:23232/deployment-operator.git
```
ssh:
```
git clone ssh://localhost:23231/deployment-operator.git
```

also, can verify that the repos are hosted in the `git-server` by running
```
ssh -p 23231 localhost
```
which should open up the TUI and show the `deployment-operator` and `scaffolds` repos.

## Checklist

### DIFF
--- a/dockerfiles/Dockerfile.softserve
+++ b/dockerfiles/Dockerfile.softserve
@@ -43,6 +43,10 @@ EXPOSE 23232
 EXPOSE 23233
 EXPOSE 9418
 
+# needs git for repos to be accessible
+RUN apk update --no-cache
+RUN apk add git
+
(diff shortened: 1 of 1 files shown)


michaeljguarino (approved):
Why does this work?

kinjalh:
soft-serve needs git installed on the machine to work but the issue was that only the builder stage had git installed, not the final (aka `server`) stage.

PR #2012, opened 2025-03-14
Fully fleshed out values.yaml for console helm chart

kinjalh:
- Add missing references into values.yaml for console helm chart (defaulting to null ~ since that will allow the value to be in the chart without effectively changing anything)
- Add documentation for values in values.yaml

Wrote a CLI util in go to do the file parsing and comparison

## Checklist

### DIFF
--- a/charts/console/values.yaml
+++ b/charts/console/values.yaml
@@ -9,57 +9,102 @@ global:
 
 cloud: 
   enabled: false
+  # If set to a truthy value, mounts a volume for the certs
+  pgRootCert: ~
 
+# disableAdditionalVolumes is the flag for global.additionalVolumes
 disableAdditionalVolumes: false
 
+# replicaCount is the replica count for console deployment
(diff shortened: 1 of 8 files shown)


michaeljguarino (commented):
I think there might actually be a standard format for helm values documentation, you might want to validate this is matching it.  

Also don't include the flaky-service binary in this pr (ideally wipe it from the commit entirely).

michaeljguarino:
also looks like somehow the console helm tests are failing

kinjalh:
oh good point on the binary, will wipe it

thanks for the notes on the other 2 taking a look

michaeljguarino (commented):
This is still missing documentation for the `kas` subchart (most relevant) and ideally also for the `kubernetes-dashoard` subchart too

kinjalh:
got it will add docstrings for those

michaeljguarino (approved):

PR #2044, opened 2025-03-21
fix: Bump design system version

maciaszczykm:
## Checklist


jsladerman (approved):

jsladerman:
regenerating the lockfile seems to be causing a TS error, could probably just hold off on that for now

PR #2062, opened 2025-03-27
provider impl

kinjalh:
Implement files function in gitlab and bitbucket providers such that they provide files in the same format as the github provider.

## Test Plan
- Unit tests added in the `consumer_test.exs` test file for the gitlab and bitbucket providers
    - `mix test test/console/ai/pubsub/vector/consumer_test.exs`
- Not included in the PR, but sandbox tested using mix tasks (i.e. essentially copy over the implementation code for the `files` function into a mix task and then run it with the URL of an actual PR)

## Checklist

### DIFF
--- a/lib/console/deployments/pr/impl/bitbucket.ex
+++ b/lib/console/deployments/pr/impl/bitbucket.ex
@@ -1,7 +1,12 @@
 defmodule Console.Deployments.Pr.Impl.BitBucket do
   import Console.Deployments.Pr.Utils
+  alias Console.Deployments.Pr.File
   alias Console.Schema.{PrAutomation, PullRequest}
+  require Logger
+
   @behaviour Console.Deployments.Pr.Dispatcher
+
+  @bitbucket_api_url "https://api.bitbucket.org/2.0"
 
   defmodule Connection do
     defstruct [:host, :token]
@@ -49,9 +54,9 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
   defp pr_content(pr), do: "#{pr["source"]["branch"]["name"]}\n#{pr["title"]}\n#{pr["summary"]["raw"]}"
 
   def review(conn, %PullRequest{url: url}, body) do
-    with {:ok, group, number} <- get_pull_id(url),
+    with {:ok, org, repo, number} <- get_pull_id(url),
          {:ok, conn} <- connection(conn) do
-      case post(conn, Path.join(["/repositories", "#{URI.encode(group)}", "pullrequests", number, "comments"]), %{
+      case post(conn, Path.join(["/repositories", "#{URI.encode("#{org}/#{repo}")}", "pullrequests", number, "comments"]), %{
         content: %{
           raw: filter_ansi(body),
           markup: "markdown"
@@ -63,17 +68,66 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
     end
   end
 
-  def files(_, _), do: {:ok, []}
+  def files(scm_conn, url) do
+    with {:ok, workspace, repo, pr_id} <- get_pull_id(url),
+         {:ok, conn} <- connection(scm_conn),
+         {:ok, pr_body, diff_url, diffstat_url} <- get_pr_info(conn, workspace, repo, pr_id),
+         diff_map = get_diff_map(conn, diff_url),
+         {:ok, diff_list} <- files_diff_list(conn, diffstat_url) do
+
+      files_list = Enum.map(diff_list["values"], fn file ->
+        filename = file["new"]["escaped_path"]
+
+        %File{
+          url: url,
+          repo: pr_body["source"]["repository"]["full_name"],
+          title: pr_body["title"],
+          contents: get_file_from_raw(conn, file["new"]["links"]["self"]["href"]),
+          filename: filename,
+          sha: pr_body["source"]["commit"]["hash"],
+          patch: Map.get(diff_map, filename, "No diff found for #{filename}"),
+          base: pr_body["destination"]["branch"]["name"],
+          head: pr_body["source"]["branch"]["name"]
+        }
+      end)
+      |> Enum.filter(&File.valid?/1)
+      {:ok, files_list}
+    else
+      err ->
+        Logger.info("failed to list pr files #{inspect(err)}")
+        err
+    end
+  end
 
   defp post(conn, url, body) do
     HTTPoison.post("#{conn.host}#{url}", Jason.encode!(body), Connection.headers(conn))
     |> handle_response()
   end
 
+  defp get(%Connection{} = conn, url) do
+    HTTPoison.get("#{conn.host}#{url}", Connection.headers(conn))
+    |> handle_response()
+  end
+
+  defp get(url, headers) when is_binary(url) and is_list(headers) do
+    HTTPoison.get(url, headers)
+    |> handle_response()
+  end
+
+  defp get_raw(url, headers) when is_binary(url) and is_list(headers) do
+    HTTPoison.get(url, headers)
+    |> handle_response_raw()
+  end
+
   defp handle_response({:ok, %HTTPoison.Response{status_code: code, body: body}})
     when code >= 200 and code < 300, do: Jason.decode(body)
   defp handle_response({:ok, %HTTPoison.Response{body: body}}), do: {:error, "bitbucket request failed: #{body}"}
   defp handle_response(_), do: {:error, "unknown bitbucket error"}
+
+  defp handle_response_raw({:ok, %HTTPoison.Response{status_code: code, body: body}})
+    when code >= 200 and code < 300, do: body
+  defp handle_response_raw({:ok, %HTTPoison.Response{body: body}}), do: {:error, "bitbucket request failed: #{body}"}
+  defp handle_response_raw(_), do: {:error, "unknown bitbucket error"}
 
   defp state(%{"state" => "MERGED"}), do: :merged
   defp state(%{"state" => "DECLINED"}), do: :closed
@@ -84,16 +138,69 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
   defp owner(_), do: nil
 
   defp connection(conn) do
-    with {:ok, url, token} <- url_and_token(conn, "https://api.bitbucket.org/2.0"),
+    with {:ok, url, token} <- url_and_token(conn, @bitbucket_api_url),
       do: Connection.new(url, token)
   end
 
   defp get_pull_id(url) do
-    with {:ok, %URI{path: "/" <> path}} <- URI.parse(url),
-         [org, repo, "pull-requests" <> number] <- String.split(path, "/") do
-      {:ok, "#{org}/#{repo}", number}
+    with %URI{path: "/" <> path} <- URI.parse(url),
+         parts <- String.split(path, "/"),
+         [workspace, repo, "pull-requests", pr_id] <- parts do
+      {:ok, workspace, repo, pr_id}
     else
       _ -> {:error, "could not parse bitbucket url"}
     end
+  end
+
+  defp get_pr_info(conn, workspace, repo, pr_id) do
+    with {:ok, %{
+           "links" => %{
+             "diff" => %{"href" => diff_url},
+             "diffstat" => %{"href" => diffstat_url}
+           }
+         } = pr_body} <- get(conn, "/repositories/#{workspace}/#{repo}/pullrequests/#{pr_id}") do
+      {:ok, pr_body, diff_url, diffstat_url}
+    else
+      {:error, %HTTPoison.Error{reason: reason}} ->
+        {:error, "HTTP request failed: #{inspect(reason)}"}
+      {:error, %Jason.DecodeError{}} ->
+        {:error, "Invalid JSON response"}
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+  defp files_diff_list(conn, diffstat_url) do
+    get(diffstat_url, Connection.headers(conn))
+  end
+
+  defp get_file_from_raw(conn, url) do
+    get_raw(url, Connection.headers(conn))
+  end
+
+  defp get_diff_map(conn, diff_url) do
+    get_raw(diff_url, Connection.headers(conn))
+    |> parse_diff_to_map()
+  end
+
+  defp parse_diff_to_map(diff_content) do
+    # Split the diff content by file
+    diff_content
+    |> String.split(~r/diff --git a\//)
+    |> Enum.reject(&(&1 == ""))
+    |> Enum.map(fn file_diff ->
+      # Extract filename and diff content
+      case Regex.run(~r/^(.+?) b\/(.+?)(?:\n|$)/, file_diff) do
+        [_, _, filename] ->
+          {filename, "diff --git a/" <> file_diff}
+        _ ->
+          # Handle new files differently
+          case Regex.run(~r/^(.+?) b\/(.+?)(?:\n|$)/, "diff --git a/" <> file_diff) do
+            [_, _, filename] -> {filename, "diff --git a/" <> file_diff}
+            _ -> {"unknown", file_diff}
+          end
+      end
+    end)
+    |> Map.new()
   end
 end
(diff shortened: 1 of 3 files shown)


lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -63,7 +63,85 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
...
+      case HTTPoison.get("#{conn.host}/repositories/#{URI.encode(group)}/pullrequests/#{number}", Connection.headers(conn)) do
+        {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
+          case Jason.decode(body) do
+            {:ok, pr} ->
+              # Then get the diff to get file changes
+              case HTTPoison.get("#{conn.host}/repositories/#{URI.encode(group)}/pullrequests/#{number}/diff", Connection.headers(conn)) do
+                {:ok, %HTTPoison.Response{status_code: 200, body: diff}} ->
+                  changes = parse_diff(diff)
+                  files = Enum.map(changes, fn change ->
+                    _file_status = cond do
```
michaeljguarino, line 79: rm these (also usually a case is better than a cond, in this case it would be:

```elixir
case change do
%{"status" => "added"} -> "(added)"
...
end
```

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -63,7 +63,85 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
...
+                      change["status"] == "removed" -> "(deleted)"
+                      change["status"] == "renamed" -> "(renamed)"
+                      true -> "(modified)"
+                    end
+
+                    %Console.Deployments.Pr.File{
+                      url: url,
+                      repo: get_repo_url(url),
+                      title: pr["title"],
+                      contents: change["patch"],
```
michaeljguarino, line 90: `contents` should be the full file contents

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -63,7 +63,85 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
     end
   end
 
-  def files(_, _), do: {:ok, []}
+  def files(conn, url) do
+    with {:ok, group, number} <- get_pull_id(url),
+         {:ok, conn} <- connection(conn) do
+      # First get PR details to get the title
+      case HTTPoison.get("#{conn.host}/repositories/#{URI.encode(group)}/pullrequests/#{number}", Connection.headers(conn)) do
```
michaeljguarino, line 70: this case should be done w/in the with chain, along with the `Jason.decode`

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -50,20 +50,97 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
...
+    else
+      {:error, error} -> {:error, error}
+      _ -> {:error, "unknown bitbucket error"}
+    end
+  end
+
+  def files(conn, url) do
+    with {:ok, group, number} <- get_pull_id(url),
+         {:ok, conn} <- connection(conn),
+         {:ok, %HTTPoison.Response{status_code: 200, body: body}} <- HTTPoison.get("#{conn.host}/repositories/#{URI.encode(group)}/pullrequests/#{number}", Connection.headers(conn)),
```
michaeljguarino, line 70: move the URI.encode to the top and save to a var so it doesn't have to be repeated

lib/console/deployments/pr/impl/gitlab.ex
```diff
@@ -70,7 +70,67 @@ defmodule Console.Deployments.Pr.Impl.Gitlab do
...
+         {:ok, %HTTPoison.Response{status_code: 200, body: body}} <- HTTPoison.get("#{conn.host}/api/v4/projects/#{uri_encode(group)}/merge_requests/#{number}/changes", Connection.headers(conn)),
+         {:ok, %{"changes" => changes}} <- Jason.decode(body) do
+
+      files = Enum.map(changes, fn change ->
+        # Get full file content for the new_path if it's not deleted
+        file_content = if change["deleted_file"] do
+          ""
+        else
+          # Get the full file content from the repository files API
+          case get_file_content(conn, group, change["new_path"], "HEAD") do
```
michaeljguarino, line 85: I'd just have these functions take the fallback as the final arg and then return them out of the internal case statement so you don't have to wrap it in this func

michaeljguarino (commented):
Could we add mocked tests for these?  I think i have one already for github that can be imitated and the logic is relatively complex.

kinjalh:
regarding a mocked test i'll try and write one up, I was running into some weird issues running `mix test` (even without making any changes myself) so will see if i can fix that locally.

I did test the files code by copying the functionality into a mix task and running locally (seems to work) but yeah fair unit test makes the most sense.

rest of the comments should be able to address on revision

michaeljguarino:
> regarding a mocked test i'll try and write one up, I was running into some weird issues running `mix test` (even without making any changes myself) so will see if i can fix that locally.
> 
> I did test the files code by copying the functionality into a mix task and running locally (seems to work) but yeah fair unit test makes the most sense.
> 
> rest of the comments should be able to address on revision

I recently added an e2e test w/ MCP that has a docker compose change, so you might need to rerun `make testup` to get the full environment working.  It'll emit weird log messages now because that's just what that server does (working on getting another one for us to use)

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -63,7 +66,39 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
     end
   end
 
-  def files(_, _), do: {:ok, []}
+  def files(_, url) do
+    case get_pr_info(url) do
```
michaeljguarino, line 70: use a with comprehension instead of nested cases

lib/console/deployments/pr/impl/gitlab.ex
```diff
@@ -70,7 +73,51 @@ defmodule Console.Deployments.Pr.Impl.Gitlab do
     end
   end
 
-  def files(_, _), do: {:ok, []}
+  def files(_, url) do
+    case get_mr_info(url) do
+      {:ok, mr_info} ->
+        res_list = mr_info["changes"]
```
michaeljguarino, line 79: make this formatting use:

```elixir
res_list = Enum.map(mr_info["changes"], fn change -> ... end)
{:ok, res_list}
```

lib/console/deployments/pr/impl/gitlab.ex
```diff
@@ -106,4 +153,35 @@ defmodule Console.Deployments.Pr.Impl.Gitlab do
...
+    end
+  end
+
+  # Construct the raw file URL from the base URL and change info
+  defp get_raw_url(base_url, sha, change) do
+    # Extract the project path from the base URL
+    [_, project_path] = Regex.run(~r{gitlab\.com/(.+)/-/merge}, base_url)
+
+    # Construct the raw URL using the project path and file path
+    "https://gitlab.com/#{project_path}/-/raw/#{sha}/#{change["new_path"]}"
```
michaeljguarino, line 170: almost all gitlabs will be self-hosted so can't actually use gitlab.com.

Also I'm pretty sure the raw urls won't resolve since they'll likely need a session cookie.  Ideally there's a gitlab api to get file contents.

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -96,4 +131,77 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
...
+
+  defp parse_bitbucket_url(url) do
+    case Regex.run(~r{bitbucket\.org/([^/]+)/([^/]+)/pull-requests/(\d+)}, url) do
+      [_, workspace, repo_slug, pr_id] -> {:ok, workspace, repo_slug, pr_id}
+      _ -> {:error, "Invalid Bitbucket pull request URL"}
+    end
+  end
+
+  defp get_pr_url(workspace, repo_slug, pr_id) do
+    "#{@bitbucket_api_url}/repositories/#{workspace}/#{repo_slug}/pullrequests/#{pr_id}"
```
michaeljguarino, line 166: this api url needs to come from the conn, not static config since most bitbuckets are self-hosted

lib/console/deployments/pr/impl/gitlab.ex
kinjalh: Verified in sandbox testing that the links to get the files in bitbucket are API links and not just web links (ex. `"https://api.bitbucket.org/2.0/repositories/kinjal_test_ws/test_repo/src/bcacd23961feaa13b1e3d8a55ef62256a979467a/dir/myfile.txt"`)

lib/console/deployments/pr/impl/bitbucket.ex
kinjalh: addressed in revision

lib/console/deployments/pr/impl/bitbucket.ex
kinjalh: addressed in revision

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -63,7 +69,37 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
     end
   end
 
-  def files(_, _), do: {:ok, []}
+  def files(conn, url) do
+    with {:ok, workspace, repo, pr_id} <- get_pull_id(url),
+         {:ok, pr_info} <- get_pr_info(conn, workspace, repo, pr_id),
+         diff_url = pr_info["links"]["diff"]["href"],
```
michaeljguarino, line 75: I'd honestly extract these in the match at line 74 too (would technically be safer as well, eg:

```elixir
{:ok, %{"links" => %{"diff" => %{"href" => diff}, "diffstat" => %{"href" => diffstat}}} <- ...
```

(or better yet, do that in line within the function body of `get_pr_info` and return a tuple like {:ok, info, diff_url, diffstat_url}

lib/console/deployments/pr/impl/gitlab.ex
```diff
@@ -96,14 +156,42 @@ defmodule Console.Deployments.Pr.Impl.Gitlab do
...
+    end
+  end
+
+  defp get_file_contents_from_commit(conn, project_id, file_path, sha) do
+    with {:ok, url, token} <- url_and_token(conn, @gitlab_api_url),
+         api_url = extract_api_url(url),
+         encoded_project_id = URI.encode_www_form(project_id),
+         encoded_file_path = URI.encode_www_form(file_path),
+         url = "#{api_url}/projects/#{encoded_project_id}/repository/files/#{encoded_file_path}?ref=#{sha}",
+         {:ok, response} <- HTTPoison.get(url, [{"PRIVATE-TOKEN", token}, {"Content-Type", "application/json"}]),
```
michaeljguarino, line 187: would be able to call the get function here too

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -89,11 +125,84 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
...
+      false ->
+        @bitbucket_api_url
+    end
+  end
+
+  defp get_pr_info(conn, workspace, repo, pr_id) do
+    with {:ok, api_url, token} <- url_and_token(conn, @bitbucket_api_url),
+         api_url = extract_api_url(api_url),
+         pr_url = "#{api_url}/repositories/#{workspace}/#{repo}/pullrequests/#{pr_id}",
+         {:ok, pr_response} <- HTTPoison.get(pr_url, [{"Authorization", "Bearer #{token}"}, {"Content-Type", "application/json"}]),
```
michaeljguarino, line 150: imitate the style of the existing post call for gets, and encapsulate auth w/in

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -89,11 +125,84 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
...
+        {:error, "HTTP request failed: #{inspect(reason)}"}
+      {:error, %Jason.DecodeError{}} ->
+        {:error, "Invalid JSON response"}
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+  defp files_diff_list(diffstat_url) do
+    case HTTPoison.get(diffstat_url) do
```
michaeljguarino, line 164: I'm like 99% sure this fetch will need auth, likely use the encapsulated get function for it (true for other fetches too)

lib/console/deployments/pr/impl/gitlab.ex
```diff
@@ -92,18 +130,40 @@ defmodule Console.Deployments.Pr.Impl.Gitlab do
...
       do: Connection.new(url, token)
   end
 
+  defp uri_encode(str), do: URI.encode(str, & &1 != ?/ and URI.char_unescaped?(&1))
+
   defp get_pull_id(url) do
-    with {:ok, %URI{path: "/" <> path}} <- URI.parse(url),
-         [group, "/merge_requests/" <> number] <- String.split(path, "-") do
-      {:ok, group, number}
+    with %URI{path: "/" <> path} <- URI.parse(url),
```
michaeljguarino, line 140: this is going to cause a regression on the `review` function at least, maybe a few others

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -84,16 +139,66 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
...
-    with {:ok, url, token} <- url_and_token(conn, "https://api.bitbucket.org/2.0"),
+    with {:ok, url, token} <- url_and_token(conn, @bitbucket_api_url),
       do: Connection.new(url, token)
   end
 
   defp get_pull_id(url) do
-    with {:ok, %URI{path: "/" <> path}} <- URI.parse(url),
-         [org, repo, "pull-requests" <> number] <- String.split(path, "/") do
-      {:ok, "#{org}/#{repo}", number}
+    with %URI{path: "/" <> path} <- URI.parse(url),
```
michaeljguarino, line 147: same, this is going to cause a regression

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -49,7 +54,8 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
   defp pr_content(pr), do: "#{pr["source"]["branch"]["name"]}\n#{pr["title"]}\n#{pr["summary"]["raw"]}"
 
   def review(conn, %PullRequest{url: url}, body) do
-    with {:ok, group, number} <- get_pull_id(url),
+    with {:ok, org, repo, number} <- get_pull_id(url),
+         group = "#{org}/#{repo}",
          {:ok, conn} <- connection(conn) do
       case post(conn, Path.join(["/repositories", "#{URI.encode(group)}", "pullrequests", number, "comments"]), %{
```
michaeljguarino, line 60: just do `URI.encode("#{org}/#{group}")` inline

lib/console/deployments/pr/impl/gitlab.ex
```diff
@@ -59,7 +64,8 @@ defmodule Console.Deployments.Pr.Impl.Gitlab do
   def pr(_), do: :ignore
 
   def review(conn, %PullRequest{url: url}, body) do
-    with {:ok, group, number} <- get_pull_id(url),
+    with {:ok, owner, repo, number} <- get_pull_id(url),
+         group = "#{owner}/#{repo}",
          {:ok, conn} <- connection(conn) do
       case post(conn, Path.join(["/api/v4/projects", "#{uri_encode(group)}", "merge_requests", number]), %{
```
michaeljguarino, line 70: just put the uri_encode("#{owner}/#{repo}") inline here too

michaeljguarino (approved):

lib/console/deployments/pr/impl/bitbucket.ex
```diff
@@ -84,16 +139,66 @@ defmodule Console.Deployments.Pr.Impl.BitBucket do
...
+         [workspace, repo, "pull-requests", pr_id] <- parts do
+      {:ok, workspace, repo, pr_id}
     else
       _ -> {:error, "could not parse bitbucket url"}
     end
   end
+
+  defp get_pr_info(conn, workspace, repo, pr_id) do
+    with {:ok, pr_body} <- get(conn, "/repositories/#{workspace}/#{repo}/pullrequests/#{pr_id}"),
+         diff_url = pr_body["links"]["diff"]["href"],
```
michaeljguarino, line 158: oh wait, these should still be extracted via pattern match instead of nested gets though, since the former is way more robust to errors

PR #2106, opened 2025-04-08
fix: Cost management graph curve

maciaszczykm:
Before:         
Zrzut ekranu 2025-04-8 o 15 00 53

After:
Zrzut ekranu 2025-04-8 o 15 00 46


michaeljguarino (commented):
could you client side filter the dupe timestamps for now?  I'll try to figure out the cause but it might take a bit of time to get it all figured out.

jsladerman (approved):

maciaszczykm:
@michaeljguarino done.

PR #2115, opened 2025-04-10
feat: Add OIDC consent pages

maciaszczykm:
Zrzut ekranu 2025-04-10 o 11 12 37


michaeljguarino:
Isn't there also a login page? I might be misremembering.  it's also possible it is all done on the consent page

assets/src/components/login/OAuthConsent.tsx
```diff
@@ -0,0 +1,161 @@
+import { useLocation, useNavigate } from 'react-router-dom'
+import queryString, { ParsedQuery } from 'query-string'
+import { Button, IconFrame } from '@pluralsh/design-system'
+import { useCallback } from 'react'
+import { A, Flex, Span } from 'honorable'
```
jsladerman, line 5: try to avoid using honorable if you can, it's not actually installed in the console codebase (probably still builds here because it's bundled with the DS)

could just replace with a styled component

jsladerman (approved):

maciaszczykm:
@michaeljguarino Login page is already there: https://github.com/pluralsh/console/blob/2db4fd1ea200b0863065a9fca6b8431807f0175b/assets/src/components/login/Login.tsx#L102

If there is anything missing let me know I will update it.

assets/src/components/login/OAuthConsent.tsx
maciaszczykm: I forgot, it's done now.

PR #2120, opened 2025-04-11
fix: argo-rollouts compatibility scraper

maciaszczykm:


michaeljguarino (commented):
cool thanks for knocking that out

michaeljguarino (approved):

PR #2121, opened 2025-04-11
Use non root user for server process

kinjalh:
Run soft-serve server as non-root (https://linear.app/pluralsh/issue/PROD-3414/git-server-ideally-shouldnt-run-as-root)
Make sure that UID is set for our non-root user so that the value can be set in the console helm chart

## Test Plan
spun up container (docker build and docker run), verified that i could still clone repos from it.

## Checklist

### DIFF
--- a/dockerfiles/Dockerfile.softserve
+++ b/dockerfiles/Dockerfile.softserve
@@ -50,4 +50,9 @@ RUN apk add git
 COPY --from=builder /usr/bin/soft /usr/bin/
 COPY --from=builder /go/data /go/data
 
+RUN addgroup -g 65543 softserve
+RUN adduser -u 65543 -G softserve -D softserve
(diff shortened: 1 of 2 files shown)


dockerfiles/Dockerfile.softserve
```diff
@@ -50,4 +50,8 @@ RUN apk add git
 COPY --from=builder /usr/bin/soft /usr/bin/
 COPY --from=builder /go/data /go/data
 
+RUN adduser -D softserve
```
michaeljguarino, line 53: make sure to give this user a fixed uid (as it'll likely need to be hardwired into the pod security context in some cases

kinjalh: sure making a revision

michaeljguarino (approved):

PR #2131, opened 2025-04-14
fix: don't set owner references on InfrastructureStack

zreigz:
## Test Plan
Apply an annotation to the imported InfrastructureStack containing the service's namespace and name. This enables the service controller to reconcile whenever the stack is modified.

## Checklist

### DIFF
--- a/go/controller/internal/controller/servicedeployment_controller.go
+++ b/go/controller/internal/controller/servicedeployment_controller.go
@@ -4,8 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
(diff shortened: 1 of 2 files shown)


michaeljguarino (commented):
could we also add code to remove the owner reference if it was already present to ensure it doesn't cause issues in existing data?

zreigz:
> could we also add code to remove the owner reference if it was already present to ensure it doesn't cause issues in existing data?

done

michaeljguarino (approved):

PR #2137, opened 2025-04-16
Pagerduty Webhook Implementation

kinjalh:
Implement pagerduty webhook. Similar idea to how the grafana webhook is implemented but specific to pagerduty payload as specified here: https://developer.pagerduty.com/docs/webhooks-overview#webhook-payload

## Test Plan
Unit tests for pagerduty like the ones existing for grafana. Will do E2E test after PR is merged.

## Checklist


lib/console/deployments/observability/webhook.ex
```diff
@@ -34,6 +34,25 @@ defmodule Console.Deployments.Observability.Webhook do
     |> ok()
   end
 
+  def payload(%ObservabilityWebhook{type: :pagerduty}, %{"event" => event = %{"data" => data}} = payload) do
+    # Create structured alert
+    # Note that pagerduty doesn't have annotations like grafana does, so just use an empty map
+    # Also, pagerduty has a single event per payload, so unlike grafana we don't iterate over messages
+    [Map.merge(%{
```
michaeljguarino, line 41: write a small listify function to simplify the list building in a pipeline, eg:

```elixir
defp listify(l) when is_list(l), do: l
defp listify(v), do: [v]
```

Then you can call it in the main pipeline here and not break flow

kinjalh: addressed in revision, made a listify function and just called it in the pipeline

PR #2139, opened 2025-04-16
Datadog observability webhook

JohnBlackwell:
## Checklist

### DIFF
--- a/lib/console/deployments/observability/webhook.ex
+++ b/lib/console/deployments/observability/webhook.ex
@@ -4,7 +4,7 @@ defmodule Console.Deployments.Observability.Webhook do
   to handle provider-specific logic.
   """
   import Console.Services.Base, only: [ok: 1]
-  alias Console.Deployments.Observability.Webhook.{Grafana, Pagerduty, Raw}
+  alias Console.Deployments.Observability.Webhook.{Grafana, Datadog, Pagerduty, Raw}
   alias Console.Schema.ObservabilityWebhook
 
   @callback associations(atom, map, map) :: map
@@ -52,6 +52,32 @@ defmodule Console.Deployments.Observability.Webhook do
     |> backfill_raw()
     |> listify()
     |> ok()
+  end
+
+  def payload(%ObservabilityWebhook{type: :datadog}, payload) do
+    case Datadog.normalize_datadog_payload(payload) do
+      [] -> {:error, "invalid payload"}
+      alerts ->
+        Enum.map(alerts, fn alert ->
+          Map.merge(%{
+            type: :datadog,
+            fingerprint: Map.get(alert, "id"),
+            annotations: Map.get(alert, "meta") || %{},
+            state: Datadog.state(
+              Map.get(alert, "status") ||
+              Map.get(alert, "alert_transition") ||
+              ""
+            ),
+            severity: Datadog.severity(alert),
+            url: Map.get(alert, "link") || Map.get(alert, "url") || "",
+            title: Map.get(alert, "title") || "Datadog Alert",
+            message: Datadog.summary(alert),
+            tags: Datadog.datadog_tags(alert),
+          }, add_associations(Datadog, alert))
+          |> backfill_raw()
+        end)
+        |> ok()
+    end
   end
 
   def payload(_, _), do: {:error, "invalid payload"}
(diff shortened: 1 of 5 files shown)


lib/console/deployments/observability/webhook/datadog.ex
```diff
@@ -0,0 +1,180 @@
...
+  end
+
+  defp extract_from_text(:project, text, acc) do
+    case extract_with_regex(text, @project_regex) do
+      nil -> acc
+      val -> Map.put(acc, :project_id, project(val))
+    end
+  end
+
+  defp extract_from_text(:cluster, text, acc) do
```
michaeljguarino, line 120: all of this is already done by the raw extractor that you likely are using in the webhook module referencing this

lib/console/deployments/observability/webhook/datadog.ex
```diff
@@ -0,0 +1,180 @@
...
+  @behaviour Console.Deployments.Observability.Webhook
+  import Console.Deployments.Observability.Webhook.Base
+
+  # Regular expressions to extract information from unstructured text
+  @service_regex [~r/service[:\s]+([^,\n]+)/i, ~r/service_name[:\s]+([^,\n]+)/i]
+  @cluster_regex [~r/cluster[:\s]+([^,\n]+)/i, ~r/cluster_name[:\s]+([^,\n]+)/i]
+  @project_regex [~r/project[:\s]+([^,\n]+)/i, ~r/namespace[:\s]+([^,\n]+)/i]
+
+  # Basic associations handlers
+  def associations(:project, %{"tags" => tags}, acc) when is_list(tags) do
```
michaeljguarino, line 11: convert the tags list to a map first, then match the keys in the map internally in a helper function (will avoid the need to keep looping over that list compounding a ton of o(n) work)

lib/console/deployments/observability/webhook/datadog.ex
```diff
@@ -0,0 +1,180 @@
...
+
+  # Helper functions for Datadog payload handling
+
+  def normalize_datadog_payload(%{"alerts" => alerts}) when is_list(alerts), do: alerts
+  def normalize_datadog_payload(%{"event" => event}) when is_map(event), do: [event]
+  def normalize_datadog_payload(%{"events" => events}) when is_list(events), do: events
+  def normalize_datadog_payload(%{"alert" => alert}) when is_map(alert), do: [alert]
+  def normalize_datadog_payload(%{"series" => _} = event), do: [event] # Metric alert format
+  def normalize_datadog_payload(payload) when is_map(payload) do
+    if Map.has_key?(payload, "body") or Map.has_key?(payload, "message") or Map.has_key?(payload, "title"), do: [payload], else: []
```
michaeljguarino, line 160: this should all be pattern matches

lib/console/deployments/observability/webhook.ex
```diff
@@ -54,6 +54,32 @@ defmodule Console.Deployments.Observability.Webhook do
     |> ok()
   end
 
+  def payload(%ObservabilityWebhook{type: :datadog}, payload) do
+    case Datadog.normalize_datadog_payload(payload) do
```
michaeljguarino, line 58: normalize_datadog_payload should return an `{:ok, _} | {:error, _}` and be matched against that instead of an empty list or non-empty list, to adhere to more conventional usage.

lib/console/deployments/observability/webhook.ex
```diff
@@ -54,6 +54,32 @@ defmodule Console.Deployments.Observability.Webhook do
...
+              Map.get(alert, "alert_transition") ||
+              ""
+            ),
+            severity: Datadog.severity(alert),
+            url: Map.get(alert, "link") || Map.get(alert, "url") || "",
+            title: Map.get(alert, "title") || "Datadog Alert",
+            message: Datadog.summary(alert),
+            tags: Datadog.datadog_tags(alert),
+          }, add_associations(Datadog, alert))
+          |> backfill_raw()
```
michaeljguarino, line 77: this is doing the regex matching you're doing internally in the datadog implementation (why i suspect it's duplicate work)

lib/console/deployments/observability/webhook/datadog.ex
```diff
@@ -0,0 +1,142 @@
+defmodule Console.Deployments.Observability.Webhook.Datadog do
+  @behaviour Console.Deployments.Observability.Webhook
+  import Console.Deployments.Observability.Webhook.Base
+
+  def associations(scope, payload, acc) do
+    handle_associations(scope, payload, acc)
+  end
+
+  defp handle_associations(:project, payload, acc) do
+    cond do
```
michaeljguarino, line 10: I don't think this cond logic actually works, refactor it to use matches with multiple function heads instead (can be done for all association types, just imitate the existing grafana logic)

michaeljguarino (approved):

PR #2140, opened 2025-04-16
fix auth handling for pagerduty webhook

kinjalh:
Verify webhook signature as shown here https://developer.pagerduty.com/docs/verifying-webhook-signatures
Previously was using basic auth (same as grafana) but for pagerduty we shouldn't do that

## Test Plan
Fixed unit tests so new verification gets tested.

## Checklist

### DIFF
--- a/lib/console_web/controllers/webhook_controller.ex
+++ b/lib/console_web/controllers/webhook_controller.ex
@@ -75,9 +75,19 @@ defmodule ConsoleWeb.WebhookController do
   end
 
   defp verify(conn, %ObservabilityWebhook{type: :pagerduty, secret: secret}) do
-    with {_, password} <- Plug.BasicAuth.parse_basic_auth(conn),
-         true <- Plug.Crypto.secure_compare(secret, password) do
-      :ok
+    with [signature] <- get_req_header(conn, "x-pagerduty-signature"),
+         mac = :crypto.mac(:hmac, :sha256, secret, Enum.reverse(conn.assigns.raw_body)),
+         computed_signature = Base.encode16(mac, case: :lower) do
+
+      String.split(signature, ",")
+      |> Enum.any?(fn
+        "v1=" <> sig -> Plug.Crypto.secure_compare(sig, computed_signature)
+        _ -> false
+      end)
+      |> case do
+        true -> :ok
+        false -> :reject
+      end
     else
       _ -> :reject
     end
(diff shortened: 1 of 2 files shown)


michaeljguarino:
I don't think you're parsing the pagerduty signature header properly, according to this it'll be something like `v1={b16str..},v1={b16str...}`

kinjalh:
yeah mb, just realized. fixing it

lib/console_web/controllers/webhook_controller.ex
```diff
@@ -75,9 +75,19 @@ defmodule ConsoleWeb.WebhookController do
   end
 
   defp verify(conn, %ObservabilityWebhook{type: :pagerduty, secret: secret}) do
-    with {_, password} <- Plug.BasicAuth.parse_basic_auth(conn),
-         true <- Plug.Crypto.secure_compare(secret, password) do
-      :ok
+    with [signature] <- get_req_header(conn, "x-pagerduty-signature") do
+      # PagerDuty webhook signature: https://developer.pagerduty.com/docs/verifying-webhook-signatures
+      v1_signatures = String.split(signature, ",")
```
michaeljguarino, line 80: usually clean elixir code minimizes the number of bound variables, so this should really become something like:

```elixir
sig = {compute-hmac}

String.split(signature, ",")
|> Enum.any?(fn
  "v1=" <> sig -> Plug.Crypto.secure_compare(sig, signature)
  _ -> false
)
|> case do
  true -> :ok
  _ -> :reject
end
```

kinjalh: sure will revise in revision, i think it might be better to do something like:
```
with [signature] <- get_req_header(conn, "x-pagerduty-signature"),
         mac = :crypto.mac(:hmac, :sha256, secret, Enum.reverse(conn.assigns.raw_body)),
         computed_signature = Base.encode16(mac, case: :lower) do
      String.split(signature, ",")
      |> Enum.filter(&String.starts_with?(&1, "v1="))
      |> Enum.map(&String.replace_leading(&1, "v1=", ""))
      |> Enum.any?(&Plug.Crypto.secure_compare(&1, computed_signature))
      |> case do
        true -> :ok
        false -> :reject
      end
else
  ...
```
which is also idiomatic elixir to the best of my knowledge
since it is more clear on the specific steps (i.e. get the values by taking only the "v1=" prefixed strings, then remove the "v1=" prefix)

kinjalh: * after discussion realized that the pattern match is more efficient (no new lists created) and also pattern match is more robust than using the string functions so going with your suggested approach

michaeljguarino (approved):

PR #2156, opened 2025-04-22
feat: Compliance reports

maciaszczykm:
## Checklist

Plural Flow: console

### DIFF
--- a/assets/src/components/security/policies/CreatePolicyReportModal.tsx
+++ b/assets/src/components/security/policies/CreatePolicyReportModal.tsx
@@ -17,6 +17,7 @@ enum ReportFormat {
 }
 
 const fetchPolicyReport = (format: ReportFormat, token: string) => {
+  streamSaver.mitm = '/mitm.html'
   fetch(`/v1/compliance/report?format=${format}`, {
     method: 'GET',
     headers: { Authorization: `Bearer ${token}` },
--- a/assets/src/components/security/policies/CreatePolicyReportModal.tsx
+++ b/assets/src/components/security/policies/CreatePolicyReportModal.tsx
@@ -1,0 +1,125 @@
+import {
+  Button,
+  FormField,
+  ListBoxItem,
+  Modal,
+  Select,
+} from '@pluralsh/design-system'
+import { useCallback, useState } from 'react'
+import { useTheme } from 'styled-components'
+
+import { ModalMountTransition } from 'components/utils/ModalMountTransition'
+import { fetchToken } from '../../../helpers/auth.ts'
+
+enum ReportFormat {
+  CSV = 'csv',
+}
+
+const fetchPolicyReport = async (format: ReportFormat, token: string) => {
+  const response = await fetch(`/v1/compliance/report?format=${format}`, {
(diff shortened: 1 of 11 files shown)


assets/src/components/security/policies/CreatePolicyReportModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const response = await fetch(`/v1/compliance/report?format=${format}`, {
+    method: 'GET',
+    headers: { Authorization: `Bearer ${token}` },
+  })
+
+  if (!response.ok) {
+    throw new Error(`Failed to fetch report: ${response.statusText}`)
+  }
+
+  const blob: Blob = await response.blob()
```
michaeljguarino, line 28: is there no way to stream download this or does it always have to be a full hidden blob?  Perhaps this can help implement: https://www.npmjs.com/package/streamsaver

maciaszczykm: There's https://developer.mozilla.org/en-US/docs/Web/API/FileSystemWritableFileStream that would help a lot but it has limited availability (not supported in Safari yet). I am checking the library that you have provided, it looks like it is using MITM: https://github.com/jimmywarting/StreamSaver.js/blob/master/StreamSaver.js#L33. We can host it on our own though.

maciaszczykm: Looks like there might be some issues with changing MITM to self-hosted one.

michaeljguarino: what does that param actually do?  It looks it's an iframe url?

michaeljguarino: we should definitely self-host it (by putting it in the public folder) since it won't work in airgaps w/o

maciaszczykm: That may be problematic with this lib: https://github.com/jimmywarting/StreamSaver.js/issues/361. I have tried changing it and it didn't have any effect. There are number of opened tickets without answers. I'm not sure if we should use it but on the other hand there are not many options to choose from.

maciaszczykm: I was testing it on Safari where it uses blob as a fallback. After switching to Chrome MITM substitution seems to work.

jsladerman (approved):

assets/src/components/security/policies/CreatePolicyReportModal.tsx
michaeljguarino: Did you push the change with the mitm substitution?  I don't see how its configured in the code so assume it's not fully ready?

maciaszczykm: It's not fully ready I will push it soon.

maciaszczykm: It's done now. There can be issues with testing it locally. There are flags in Chrome to allow that. It should work on prod.

PR #2168, opened 2025-04-25
feat: improve helm chart loading state and upgrade design system

floreks:
- Added loading state indication to the helm chart form fields
- Updated design system to be able to use `Remeasure Fonts` action in the code editor to fix cursor position
- Split `Settings` -> `Repository` tab into `Git` and `Helm`. If `Git` is not available it will autoredirect to `Helm`.

It was a quick change to split `Repository` into 2 tabs. It can definitely be refactored to use shared component as base.

### Git
image

### Only Git
image

### Helm - default state
image

### Helm - Loading state
image

### Only Helm
image

## Test Plan

Locally

## Checklist

Plural Flow: console

### DIFF
--- a/assets/package.json
+++ b/assets/package.json
@@ -47,7 +47,7 @@
     "@nivo/radial-bar": "0.88.0",
     "@nivo/tooltip": "0.88.0",
     "@nivo/treemap": "0.88.0",
-    "@pluralsh/design-system": "5.11.0",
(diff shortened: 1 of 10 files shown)


michaeljguarino (commented):
Could we also split the git and helm configs into separate tabs actually? That will greatly simplify the navigability here

michaeljguarino (approved):
awesome, looks good to me

PR #2169, opened 2025-04-25
Openai to ollama ai proxy

JohnBlackwell:
## Checklist

Plural Flow: console


go/ai-proxy/Makefile
```diff
@@ -31,8 +31,8 @@ build: fmt vet ## build manager binary
 .PHONY: run
 run: fmt vet ## run a proxy locally
 	go run ./ \
-		--provider=${PROVIDER} \
-		--provider-host=${PROVIDER_API_HOST} \
+		--provider=ollama \
```
michaeljguarino, line 34: might want to revert this change

go/ai-proxy/proxy/ollama/ollama.go
```diff
@@ -0,0 +1,248 @@
...
+				if err != nil {
+					return err
+				}
+
+				toolCalls = append(toolCalls, openai.ToolCall{
+					ID:    uuid.NewString(),
+					Index: i,
+					Type:  "function",
+					Function: struct {
+						Name      string `json:"name"`
```
michaeljguarino, line 92: could we make a small type for this instead of an inline struct?

go/ai-proxy/proxy/ollama/ollama.go
```diff
@@ -0,0 +1,248 @@
...
+		klog.ErrorS(err, "failed to convert ollama request")
+		return
+	}
+
+	respFunc := func(resp ollamaapi.ChatResponse) error {
+		chunk := openai.ChunkChoice{}
+		var finishReason *string
+
+		// Process any tool calls in the response
+		if len(resp.Message.ToolCalls) > 0 {
```
michaeljguarino, line 78: do we know the tool calls here have stable indexes?  That's important to be able to build the full json response on the client while streaming

go/ai-proxy/proxy/ollama/ollama.go
```diff
@@ -0,0 +1,248 @@
...
+			var toolCalls []openai.ToolCall
+			for i, tc := range resp.Message.ToolCalls {
+				// Convert the map back to a JSON string for arguments
+				argBytes, err := json.Marshal(tc.Function.Arguments)
+				if err != nil {
+					return err
+				}
+
+				toolCalls = append(toolCalls, openai.ToolCall{
```
michaeljguarino, line 87: openai also has a callId that's relatively important, not sure if ollama tries to provide it

go/ai-proxy/proxy/ollama/ollama.go
JohnBlackwell: it's actually a field in the ollama go lib for their openai integration https://github.com/ollama/ollama/blob/main/openai/openai.go#L156

go/ai-proxy/proxy/ollama/ollama.go
JohnBlackwell: I'm not 100% sure either way. I updated this to use the ollama toolfunction index field but I can't say for certain that they're stable.

go/ai-proxy/proxy/ollama/ollama.go
JohnBlackwell: they don't :(

go/ai-proxy/proxy/ollama/ollama.go
```diff
@@ -0,0 +1,314 @@
...
+	return &ToolRegistry{
+		tools:       make(map[string]ollamaapi.ToolCall),
+		toolIndices: make(map[string]int),
+		toolCallIds: make(map[string]string),
+		nextIndex:   0,
+	}
+}
+
+func toolCallId() string {
+	const letterBytes = "abcdefghijklmnopqrstuvwxyz0123456789"
```
michaeljguarino, line 49: can just use polly's impl here: https://github.com/pluralsh/polly/blob/main/algorithms/random.go#L59

go/ai-proxy/proxy/ollama/ollama.go
```diff
@@ -0,0 +1,314 @@
...
+
+type OllamaProxy struct {
+	toolRegistry *ToolRegistry
+	ollamaClient *ollamaapi.Client
+	api.OpenAIProxy
+}
+
+// ToolRegistry manages a collection of tools for efficient lookup
+type ToolRegistry struct {
+	tools       map[string]ollamaapi.ToolCall
```
michaeljguarino, line 33: instead of having three hash maps, you should probably make an entry struct with three fields like `ToolEntry` and make the registry store a map[string]ToolEntry (that way you don't have to defensively program around the key being in one map vs the others).

go/ai-proxy/proxy/ollama/ollama.go
```diff
@@ -0,0 +1,319 @@
...
+func (r *toolRegistry) Register(tool ollamaapi.ToolCall) {
+	name := tool.Function.Name
+	if _, exists := r.entries[name]; !exists {
+		r.entries[name] = toolEntry{
+			tool:   tool,
+			index:  r.nextIndex,
+			callId: algorithms.String(8),
+		}
+		r.nextIndex++
+	} else {
```
michaeljguarino, line 57: why does this else condition need to fire?  Can't you just ignore it

go/ai-proxy/proxy/ollama/ollama.go
```diff
@@ -0,0 +1,319 @@
...
+func (o *OllamaProxy) convertOpenAIToOllamaChatRequest(req *openai.ChatCompletionRequest) (*ollamaapi.ChatRequest, error) {
+	var messages []ollamaapi.Message
+
+	for _, tool := range req.Tools {
+		otc := ollamaapi.ToolCall{
+			Function: ollamaapi.ToolCallFunction{
+				Name: tool.Function.Name,
+			},
+		}
+		if _, _, _, exists := o.toolRegistry.Get(otc.Function.Name); !exists {
```
michaeljguarino, line 301: I think here you can just call `o.toolRegistry.Register` since it does the existence check internally

michaeljguarino (approved):

PR #2201, opened 2025-05-05
feat: add ClusterSync resource

zreigz:
Template example:
```
apiVersion: deployments.plural.sh/v1alpha1
kind: ClusterSync
metadata:
  name: clustersync-sample
spec:
  clusterSpec:
    metadata:
      name: "{{ .cluster.handle }}"
    spec:
      handle: "{{ .cluster.handle }}"
      version: "{{ if .cluster.version }}{{ .cluster.version }}{{ else }}null{{ end }}"
      protect: "{{ if .cluster.protect }}{{ .cluster.protect }}{{ else }}false{{ end }}"
      tags: "{ {{- range $i, $t := .cluster.tags }}{{ if $i }}, {{ end }}{{ $t.name }}: {{ $t.value }}{{- end }} }"
      metadata: "{{ .cluster.metadata | toYaml | nindent 8}}"
      bindings:
        read: '{{ if .cluster.readBindings }}{{ range .cluster.readBindings }}{{ printf "\n  - UserID: %s" .user.id | nindent 2 }}{{ end }}{{ else }}null{{ end }}'
        write: '{{ if .cluster.writeBindings }}{{ range .cluster.writeBindings }}{{ printf "\n  - UserID: %s" .user.id | nindent 2 }}{{ end }}{{ else }}null{{ end }}'
```

**CONSTRAINT:** Since all templated fields are strings, they must be written on a single line to ensure proper unmarshaling. The controller must remove the surrounding quotation marks before unmarshaling to preserve the intended type.

## Checklist

Plural Flow: console


michaeljguarino (commented):
We should also fetch crds created by a cluster sync that are no longer in the cluster gql api list and delete the from k8s in this controller if possible.  I think an algo like:

* add a default label so they can be listed easily, before each reconcile list all created cluster crs across all namespaces
* put them into a set, and remove from the set as you create
* any crs in the set after the run should be deleted

zreigz:
> We should also fetch crds created by a cluster sync that are no longer in the cluster gql api list and delete the from k8s in this controller if possible. I think an algo like:
> 
> * add a default label so they can be listed easily, before each reconcile list all created cluster crs across all namespaces
> * put them into a set, and remove from the set as you create
> * any crs in the set after the run should be deleted

good point

go/controller/internal/controller/clustersync_controller.go
```diff
@@ -0,0 +1,224 @@
...
+			if err != nil {
+				allErrs = append(allErrs, err)
+				logger.Error(err, "failed to template cluster")
+				continue
+			}
+			if clusterCRD.Namespace == "" {
+				clusterCRD.Namespace = clusterSync.Namespace
+			}
+			existingCluster := &v1alpha1.Cluster{}
+			if err := r.Client.Get(ctx, types.NamespacedName{Name: clusterCRD.Name, Namespace: clusterCRD.Namespace}, existingCluster); err != nil {
```
michaeljguarino, line 101: also this is a pretty big function, i think we should refactor the actual crud of the cr into a separate function, and leave this to just handle pagination and filtering out dangling clusters to be pruned at the end.

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):

PR #2207, opened 2025-05-06
linkerd impl + ebpf unit test

kinjalh:
implementation of Linkerd service mesh.
Also added tests for EBPF service mesh since it didn't have a unit test file.

## Test Plan
mix test

## Checklist

Plural Flow: console

### DIFF
--- a/lib/console/mesh/provider/linkerd.ex
+++ b/lib/console/mesh/provider/linkerd.ex
@@ -1,0 +1,69 @@
+defmodule Console.Mesh.Provider.Linkerd do
+  @moduledoc """
+  Implementation for linkerd, for the most part all this data is actually accessible via the prometheus.
+  Just extract it and use the graph builder module to gradually build the graph since the metrics are
+  spread across separate prometheus metrics entirely.
+  """
+
+  @behaviour Console.Mesh.Provider
+  alias Console.Mesh.Provider.Utils
+  alias Console.Schema.{Cluster}
+  alias Console.Mesh.{Edge, Workload, Statistics}
+  alias Prometheus.{Result}
+
+  require Logger
+
+  defstruct [:prom, :cluster]
+
+  # Reference: https://linkerd.io/2.12/reference/proxy-metrics/
+  @queries [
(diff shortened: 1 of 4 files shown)


lib/console/mesh/provider/linkerd.ex
```diff
@@ -0,0 +1,68 @@
...
+  alias Console.Schema.{Cluster}
+  alias Console.Mesh.{Edge, Workload, Statistics}
+  alias Prometheus.{Result}
+
+  require Logger
+
+  defstruct [:prom, :cluster]
+
+  @queries [
+    requests: ~s/rate(request_total{direction="inbound",cluster="$cluster"$additional}[5m])/,
```
michaeljguarino, line 19: can you link to the docs that specify these metrics for linkerd?  I'd assume they'd have a `linkerd_` prefix if it's emitting them

test/console/mesh/provider/linkerd_test.exs
```diff
@@ -0,0 +1,42 @@
...
+  alias Console.Mesh.{Provider, Edge}
+  alias Prometheus.{Response, Data, Result}
+  use Mimic
+
+  describe "#graph/2" do
+    test "it can fetch the graph for a cluster" do
+      cluster = insert(:cluster, operational_layout: %{service_mesh: :linkerd})
+      deployment_settings(prometheus_connection: %{host: "https://prom.example.com"})
+
+      Enum.each(1..7, fn _ ->
```
michaeljguarino, line 12: you can actually use expect(Console.Mesh.Prometheus, :query, 7, ...)

kinjalh: sounds good will address in revision

lib/console/mesh/provider/linkerd.ex
kinjalh: https://linkerd.io/2.12/reference/proxy-metrics/ - i don't believe linkerd metrics are prefixed, the docs just have the raw metric name. (for comparison, it looks like istio https://istio.io/latest/docs/reference/config/metrics/ does have metrics prefixed)

michaeljguarino (approved):

PR #2215, opened 2025-05-07
cilium service mesh impl + unit test

kinjalh:
similar to other service mesh impls
https://linear.app/pluralsh/issue/PROD-3343/add-support-for-cilium

## Test Plan
unit tests

## Checklist

Plural Flow: console

### DIFF
--- a/lib/console/mesh/provider/cilium.ex
+++ b/lib/console/mesh/provider/cilium.ex
@@ -1,0 +1,65 @@
+defmodule Console.Mesh.Provider.Cilium do
+  @moduledoc """
+  A provider for Cilium.
+  """
+
+  @behaviour Console.Mesh.Provider
+  alias Console.Mesh.Provider.Utils
+  alias Console.Schema.{Cluster}
+  alias Console.Mesh.{Edge, Workload, Statistics}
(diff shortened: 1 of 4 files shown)


lib/console/mesh/provider/cilium.ex
```diff
@@ -0,0 +1,64 @@
...
+  alias Console.Schema.{Cluster}
+  alias Console.Mesh.{Edge, Workload, Statistics}
+  alias Prometheus.{Result}
+
+  require Logger
+
+  defstruct [:prom, :cluster]
+
+  @queries [
+    connections: ~s/sum(hubble_flows_processed_total{cluster="$cluster"$additional})/,
```
michaeljguarino, line 17: I didn't realize this initially but these keys need to map to the fields in the Console.Mesh.Statistics struct (pretty sure this is significantly different)

kinjalh: oh ok. i'll make the changes then for this one and also for the other mesh impls if they don't match up

michaeljguarino:
we should probably add test coverage to ensure the correct fields are set too btw

PR #2234, opened 2025-05-13
feat: Respect content-disposition header

maciaszczykm:
Did not notice it before as on localhost headers are limited due for CORS requests.

### DIFF
--- a/assets/src/components/security/policies/CreatePolicyReportModal.tsx
+++ b/assets/src/components/security/policies/CreatePolicyReportModal.tsx
@@ -22,7 +22,14 @@ const fetchPolicyReport = (format: ReportFormat, token: string) => {
     method: 'GET',
     headers: { Authorization: `Bearer ${token}` },
   }).then((res) => {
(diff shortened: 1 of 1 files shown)


assets/src/components/security/policies/CreatePolicyReportModal.tsx
```diff
@@ -22,7 +22,15 @@ const fetchPolicyReport = (format: ReportFormat, token: string) => {
     method: 'GET',
     headers: { Authorization: `Bearer ${token}` },
   }).then((res) => {
-    const writeStream = streamSaver.createWriteStream('report.zip')
+    const filename =
```
michaeljguarino, line 25: is there maybe a npm package or pre-made impl for this? At least to parse the headers first then maybe you can parse the content disposition from there

maciaszczykm: There is https://github.com/jshttp/content-disposition. I will use it.

michaeljguarino (approved):

PR #2240, opened 2025-05-13
Add opensearch support (vector store)

kinjalh:
add support for AWS Opensearch as a vector store option. In comparison to the existing Elasticsearch vector store impl:

- Use AWS Sig V4 (Supported by Req)
- Some differences in format between Elasticsearch and Opensearch payloads
- Use sigv4-proxy, localstack, and opensearch docker images to simulate a test AWS Opensearch cluster locally

Also some small fixes in existing Elasticsearch implementation, and add a fetch test for it as well.

## Test Plan
- Set up a "mock" AWS Opensearch local cluster using sigv4-proxy, localstack, and opensearch image
- Add test to fetch PR file

## Checklist

Plural Flow: console


lib/console/logs/provider/opensearch.ex
```diff
@@ -0,0 +1,197 @@
...
+        log: source["message"],
+        timestamp: Timex.parse!(source["@timestamp"], "{ISO:Extended:Z}"),
+        facets: facets(source)
+      }
+    end)
+    |> Enum.filter(& &1.log)
+  end
+  defp format_hits(_), do: []
+
+  defp build_query(%Query{query: str} = q) do
```
michaeljguarino, line 58: I'm guessing all this elasticsearch logic is shared right?  We should probably refactor it into its own module, like ElasticSearchUtils or something

lib/console/ai/vector/opensearch.ex
```diff
@@ -0,0 +1,163 @@
...
+  defp query_filters(query, _), do: query
+
+  defp doc_filters(doc, [_ | _] = filters) do
+    Map.put(doc, :filters, Map.new(filters))
+  end
+  defp doc_filters(doc, _), do: doc
+
+  defp url(%Opensearch{host: host}, path), do: Path.join(host, path)
+
+  defp headers(%Opensearch{} = os) do
```
michaeljguarino, line 151: We should probably move these to a util too

lib/console/schema/deployment_settings.ex
```diff
@@ -40,6 +40,25 @@ defmodule Console.Schema.DeploymentSettings do
     end
   end
 
+  defmodule Opensearch do
```
michaeljguarino, line 43: can put the headers extraction here

lib/console/ai/vector/opensearch.ex
kinjalh: moving them to `lib/console/ai/vector/utils.ex`
will also refactor the common stuff in the log driver impls

lib/console/logs/provider/opensearch.ex
kinjalh: the `build_query` syntax is identical but the `add_terms`, `add_namespaces`, `add_facets`, etc are different for each provider since opensearch and elastic have some differences in query format. So I think this is probably better to not refactor

lib/console/schema/deployment_settings.ex
kinjalh: yeah after poking around a bit it seems best to do everything here. I'm doing this for the Elastic (both vectorstore and logs provider), Opensearch (both vectorstore and logs provider), and Connection (used by `lib/console/logs/provider/victoria.ex`)

lib/console/logs/provider/opensearch.ex
michaeljguarino: Ok that makes sense then

lib/console/schema/deployment_settings.ex
```diff
@@ -5,7 +5,14 @@ defmodule Console.Schema.DeploymentSettings do
 
   defenum AIProvider, openai: 0, anthropic: 1, ollama: 2, azure: 3, bedrock: 4, vertex: 5
   defenum LogDriver, victoria: 0, elastic: 1
-  defenum VectorStore, elastic: 0
+  defenum VectorStore, elastic: 0, opensearch: 1
+
+  @json_header [{"Content-Type", "application/json"}]
```
michaeljguarino, line 10: oh i don't think you necessarily need to move those there (I mean you can, but it's not really deployment settings-y)

lib/console/schema/deployment_settings.ex
```diff
@@ -21,6 +28,15 @@ defmodule Console.Schema.DeploymentSettings do
       |> cast(attrs, ~w(host user password)a)
       |> validate_required([:host])
     end
+
+    def to_headers(%__MODULE__{user: u, password: p}) when is_binary(u) and is_binary(p) do
```
michaeljguarino, line 32: i'd just name these functions `headers` and `url` (the to is kind of redundant and rather save the characters)

lib/console/schema/deployment_settings.ex
kinjalh: i can parametrize it and have the provider impls pass it in

lib/console/schema/deployment_settings.ex
```diff
@@ -308,8 +368,10 @@ defmodule Console.Schema.DeploymentSettings do
   end
 
   defp set_initialized(changeset) do
-    case {get_change(changeset, :store), get_change(changeset, :elastic)} do
-      {:elastic, e} when not is_nil(e) ->
+    case {get_change(changeset, :store), get_change(changeset, :elastic), get_change(changeset, :opensearch)} do
+      {:elastic, e, _} when not is_nil(e) ->
+        put_change(changeset, :initialized, false)
+      {:opensearch, o, _} when not is_nil(o) ->
```
michaeljguarino, line 374: this needs to be `{:opensearch, _, o} when not is_nil(o)`

michaeljguarino (commented):
there also is going to need to be a graphql schema change to support this btw

lib/console/schema/deployment_settings.ex
kinjalh: addressing this + graphql stuff in revision

lib/console/graphql/deployments/settings.ex
```diff
@@ -68,6 +69,15 @@ defmodule Console.GraphQl.Deployments.Settings do
     field :password, :string
   end
 
+  input_object :opensearch_connection_attributes do
+    field :host,     non_null(:string)
+    field :index,    non_null(:string)
+    field :aws_access_key_id, :string
+    field :aws_secret_access_key, :string
+    field :aws_region, :string
+    field :aws_session_token, :string
```
michaeljguarino, line 78: you probably wouldn't ever manually specify session token (it expires).

Also make sure these are vertically aligned

kinjalh: sure will remove

test/console/ai/pubsub/vector/consumer_test.exs
```diff
@@ -186,16 +187,119 @@ defmodule Console.AI.PubSub.Vector.ConsumerTest do
...
-      {:ok, c} = count_index(vector_index())
+      {:ok, c} = ES.count_index(ES.vector_index())
       assert c > 0
 
       settings = Console.Deployments.Settings.fetch_consistent()
       assert settings.ai.vector_store.initialized
     end
   end
 
+  describe "FetchPRFiles" do
```
michaeljguarino, line 200: minor nit but the way these tests were organized is each describe block matches the event its handling (which i suspect this doesn't map to), you could just put them under the other describe for the event itself

michaeljguarino (approved):

test/console/ai/pubsub/vector/consumer_test.exs
kinjalh: ah ok i'll move these to the PullRequestCreated block since both these new tests do first create a PR

PR #2261, opened 2025-05-19
AI Chat Tool for Alerts

kinjalh:
Implement tool to fetch Alerts for AI chat. Allow filtering by type (newrelic, pagerduty, etc), severity (critical, low, etc), and state (firing, resolved).

Also minor modification to `Alert.for_flow` where we add the `distinct: true` condition to make sure we do not get duplicates.

## Test Plan
Unit test

## Checklist

Plural Flow: console

### DIFF
--- a/lib/console/ai/tools/alerts.ex
+++ b/lib/console/ai/tools/alerts.ex
@@ -1,0 +1,51 @@
+defmodule Console.AI.Tools.Alerts do
+  use Ecto.Schema
+  import Ecto.Changeset
+  import Console.AI.Tools.Utils
+  alias Console.Schema.{Alert, Flow}
+
+  embedded_schema do
+    field :severities, {:array, :string}
+    field :state,      :string
+    field :types,       {:array, :string}
+  end
+
+  @valid ~w(severities state types)a
+
+  def changeset(model, attrs) do
+    model
+    |> cast(attrs, @valid)
+  end
+
+  @json_schema Console.priv_file!("tools/alerts.json") |> Jason.decode!()
+
+  def json_schema(), do: @json_schema
+  def name(), do: plrl_tool("alerts")
+  def description(), do: "Shows alerts for the current flow, with optional filtering by severity, state, and type."
+
+  def implement(%__MODULE__{severities: severities, state: state, types: types}) do
+    case Console.AI.Tool.flow() do
+      %Flow{id: flow_id} ->
+        Alert.for_flow(flow_id)
+        |> apply_filters(%{severities: severities, state: state, types: types})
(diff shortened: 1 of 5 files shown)


lib/console/ai/tools/alerts.ex
```diff
@@ -0,0 +1,85 @@
...
+      {types,      :apply_type_filter}
+    ]
+    |> Enum.reduce(base_query, fn {value, filter_fun_atom}, current_query ->
+      apply(__MODULE__, filter_fun_atom, [current_query, value])
+    end)
+  end
+
+  def apply_severity_filter(query, [_ | _] = severities) do
+    atomic_severities = Enum.map(severities, &String.to_atom/1)
+    from(q in query, where: q.severity in ^atomic_severities)
```
michaeljguarino, line 53: don't manually define the queries here, instead put them in the alert schema module and call those.  It makes it easier to reuse and organize the queries against the db.

lib/console/ai/tools/alerts.ex
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def query_builder(base_query, severities, state, types) do
+    [
+      {severities, :apply_severity_filter},
+      {state,      :apply_state_filter},
+      {types,      :apply_type_filter}
+    ]
+    |> Enum.reduce(base_query, fn {value, filter_fun_atom}, current_query ->
+      apply(__MODULE__, filter_fun_atom, [current_query, value])
```
michaeljguarino, line 47: This should probably not call `apply` directly, it's higher overhead and an anti-pattern, instead you can just make the reduce something like we do in a lot of the gql resolvers, search for functions named things like `apply_filters`

kinjalh: sure modeling it after Policy.apply_filters (`lib/console/graphql/resolvers/deployments/policy.ex`)

lib/console/ai/tools/alerts.ex
```diff
@@ -0,0 +1,64 @@
...
+      {:severities, severities}, q -> Alert.for_severities(q, severities)
+      {:state, state}, q -> Alert.for_state(q, state)
+      {:types, types}, q -> Alert.for_types(q, types)
+      _, q -> q
+    end)
+  end
+
+  defp model(alerts) do
+    Enum.map(alerts, fn alert ->
+      %{
```
michaeljguarino, line 50: fwiw you could just do `Map.take(alert, ~w(type severity state title url message annotations project_id cluster_id service_id)a)` here

michaeljguarino (approved):
mostly good, a minor style nit but it's probably ok

PR #2273, opened 2025-05-20
feat(graphql): add `uid` field to ServiceDeployment components

floreks:
## Checklist

Plural Flow: console


michaeljguarino (approved):
would you really need this in the agent?

floreks:
The UI will need to read it and the agent will set it. Without that we can't be sure which child is a direct descendant of the service component.

PR #2275, opened 2025-05-20
upload compatibility matricies to s3 bucket

kinjalh:
https://linear.app/pluralsh/issue/PROD-3598/upload-compatibility-matrices-to-s3

## Test Plan
https://github.com/pluralsh/console/actions/runs/15146927945 - test run

## Checklist

Plural Flow: console

### DIFF
--- a/.github/workflows/compatability-matrices.yaml
+++ b/.github/workflows/compatability-matrices.yaml
@@ -1,0 +1,49 @@
+name: Sync Compatibility Matrices to S3
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'static/addons/**'
+      - 'static/compatibilities/**'
+      - 'static/compatibilities.yaml'
+      - 'static/versions.yml'
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  sync-to-s3:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Configure AWS credentials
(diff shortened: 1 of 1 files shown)


.github/workflows/compatability-matrices.yaml
```diff
@@ -0,0 +1,49 @@
+name: Sync Compatibility Matrices to S3
+
+on:
+  push:
+    branches:
+      - compatability-matrix-s3-upload
```
michaeljguarino, line 6: this should be master

.github/workflows/compatability-matrices.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Sync static/addons to S3
+        run: |
+          aws s3 sync ./static/addons s3://plural-www-assets/matrices/compatability/static/addons
+        if: hashFiles('static/addons/**') != ''
+
+
+      - name: Sync static/compatibilities directory to S3
+        run: |
+          aws s3 sync ./static/compatibilities s3://plural-www-assets/matrices/compatability/static/compatibilities
+        if: hashFiles('static/compatibilities/**') != ''
```
michaeljguarino, line 39: what does hasFiles do?

.github/workflows/compatability-matrices.yaml
kinjalh: oops sry will fix on revision

.github/workflows/compatability-matrices.yaml
kinjalh: conditional to avoid triggering the s3 command if the directory or file is empty

PR #2282, opened 2025-05-22
feat: Add search to stacks state graph

maciaszczykm:


assets/src/components/stacks/state/StackStateGraph.tsx
```diff
@@ -9,55 +9,96 @@ import { ReactFlowGraph } from '../../utils/reactflow/ReactFlowGraph'
...
   [NodeType.Stage]: StackStateGraphNode,
 }
 
-function getNodesAndEdges(state: StackState) {
+const searchOptions = {
+  keys: [
+    'name',
+    {
+      name: 'configuration',
+      getFn: (r) => JSON.stringify(r?.configuration ?? {}),
```
maciaszczykm, line 25: It looks like fuse.js does not offer search on all object fields so this has to be transformed.

assets/src/components/stacks/state/StackStateGraph.tsx
```diff
@@ -9,55 +9,96 @@ import { ReactFlowGraph } from '../../utils/reactflow/ReactFlowGraph'
...
+function getNodesAndEdges(state: StackState, query: string) {
+  const resources = state.state?.filter(isNonNullable) ?? []
+
+  const filteredIds = isEmpty(query)
+    ? new Set(resources.map(({ identifier }) => identifier))
+    : new Set(
+        new Fuse(resources, searchOptions)
+          .search(query)
+          .map(({ item }) => [item.identifier, ...(item.links ?? [])])
+          .flat()
```
maciaszczykm, line 40: Similar to the cluster network graph this add not only items found by the search but also items linked directly to them.

michaeljguarino (approved):

PR #2285, opened 2025-05-23
Alerts resolution tool implementation.

kinjalh:
Since alerts resolutions are indexed in vector DB, other than matching the flow id we want to rely purely on semantic search (i.e. no additional filters).

## Checklist

Plural Flow: console

### DIFF
--- a/lib/console/ai/tools/alerts_resolutions.ex
+++ b/lib/console/ai/tools/alerts_resolutions.ex
@@ -1,0 +1,44 @@
+defmodule Console.AI.Tools.AlertsResolutions do
+  use Ecto.Schema
+  import Ecto.Changeset
+  import Console.AI.Tools.Utils
+  alias Console.AI.{Tool, VectorStore}
+  alias Console.Schema.Flow
+
+  embedded_schema do
+    field :query, :string
+  end
+
+  @valid ~w(query)a
+
+  def changeset(model, attrs) do
+    model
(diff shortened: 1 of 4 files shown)


lib/console/ai/tools/alerts_resolutions.ex
```diff
@@ -0,0 +1,45 @@
...
+  @json_schema Console.priv_file!("tools/alerts_resolutions.json") |> Jason.decode!()
+
+  def json_schema(), do: @json_schema
+  def name(), do: plrl_tool("alerts_resolutions")
+  def description(), do: "Shows alert resolutions for the current flow. Uses semantic search to find resolutions."
+
+  def implement(%__MODULE__{query: query}) do
+    with {:flow, %Flow{id: flow_id}} <- {:flow, Tool.flow()},
+         true <- VectorStore.enabled?(),
+         filters = [flow_id: flow_id, datatype: {:raw, :alert_resolution}],
```
michaeljguarino, line 28: just combine filters and fetch_opts into one line here

kinjalh: sure, addressed in revision

michaeljguarino (approved):
this looks correct but let's hold off on merging until after release.  I think we're getting to the point where there are too many tools and some sort of mechanism for the llm to select them before presenting to the api is needed

PR #2328, opened 2025-06-05
Org email suffix as env var

JohnBlackwell:
## Checklist

Plural Flow: console

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,7 +2,8 @@ import Config
 
 config :console,
   ecto_repos: [Console.Repo],
-  socket: :forge_socket
+  socket: :forge_socket,
+  org_email_suffix: System.get_env("ORG_EMAIL_SUFFIX", "")
 
 
 config :piazza_core,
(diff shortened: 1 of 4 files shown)


config/config.exs
```diff
@@ -2,7 +2,8 @@ import Config
 
 config :console,
   ecto_repos: [Console.Repo],
-  socket: :forge_socket
+  socket: :forge_socket,
+  org_email_suffix: System.get_env("ORG_EMAIL_SUFFIX", "")
```
michaeljguarino, line 6: this needs to be in `rel/runtime.exs`

lib/console/services/users.ex
```diff
@@ -189,7 +189,12 @@ defmodule Console.Services.Users do
   def bootstrap_user(_), do: {:error, "Failed to bootstrap user, likely missing email claim in oidc id token"}
 
   defp sanitize_email(email) do
-    String.replace(email, "+mottmac", "")
+    case Console.conf(:org_email_suffix) do
+      suffix when is_binary(suffix) and suffix != "" ->
```
michaeljguarino, line 193: this match should really be `when is_binary(suffix) and byte_size(suffix) > 0`

PR #2343, opened 2025-06-09
fix: Fix service deployment deletion

maciaszczykm:
Improved unit tests to cover deletion process and tested manually on kind cluster connected to plrl-dev-aws.

### DIFF
--- a/go/controller/internal/controller/servicedeployment_controller.go
+++ b/go/controller/internal/controller/servicedeployment_controller.go
@@ -523,7 +523,7 @@ func (r *ServiceReconciler) addOrRemoveFinalizer(service *v1alpha1.ServiceDeploy
 
 		// Remove service from Console API if it exists and is not read-only.
 		if exists && !service.Status.IsReadonly() {
-			if err := r.ConsoleClient.DeleteService(*service.Status.ID); err != nil {
+			if err := r.deleteService(service.Status.GetID(), service.Spec.Detach); err != nil {
 				// If it fails to delete the external dependency here, return with the error
 				// so that it can be retried.
 				utils.MarkCondition(service.SetCondition, v1alpha1.SynchronizedConditionType, v1.ConditionFalse, v1alpha1.SynchronizedConditionReasonError, err.Error())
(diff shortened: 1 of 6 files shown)


michaeljguarino (approved):
what were the edge cases this covers?

go/controller/internal/controller/servicedeployment_controller.go
```diff
@@ -496,57 +497,59 @@ func (r *ServiceReconciler) addOwnerReferences(ctx context.Context, service *v1a
 	return nil
 }
 
-func (r *ServiceReconciler) handleDelete(ctx context.Context, service *v1alpha1.ServiceDeployment) (ctrl.Result, error) {
-	log := log.FromContext(ctx)
-	if controllerutil.ContainsFinalizer(service, ServiceFinalizer) {
-		log.Info("try to delete service")
-		cluster := &v1alpha1.Cluster{}
```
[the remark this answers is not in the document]
michaeljguarino: ah gotcha

go/controller/internal/controller/servicedeployment_controller.go
```diff
@@ -143,7 +145,6 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
 			utils.MarkCondition(service.SetCondition, v1alpha1.SynchronizedConditionType, v1.ConditionFalse, v1alpha1.SynchronizedConditionReasonError, err.Error())
 			return ctrl.Result{}, err
 		}
-		controllerutil.AddFinalizer(service, ServiceFinalizer)
```
maciaszczykm, line 146: I got distracted and haven't finished second comment. There is also slight chance that finalizer was not added after resource creation. Now it's added before resource is created which should also make it more error proof.

go/controller/internal/controller/servicedeployment_controller.go
```diff
@@ -496,57 +497,59 @@ func (r *ServiceReconciler) addOwnerReferences(ctx context.Context, service *v1a
...
+
+		// Stop reconciliation as the item does no longer exist.
+		return &ctrl.Result{}
 	}
-	return ctrl.Result{}, nil
+
+	return nil
 }
 
-func (r *ServiceReconciler) deleteService(service *v1alpha1.ServiceDeployment) error {
```
maciaszczykm, line 542: Deletion was previously based on ID from CRD, but since we get object from API it's better to use it as it will always be there.

PR #2348, opened 2025-06-10
feat: Cloud query gRPC service

maciaszczykm:
- Extract is not implemented yet
- We have tested queries for all three cloud providers


go/cloud-query/api/proto/cloudquery.proto
```diff
@@ -0,0 +1,74 @@
...
+
+message AzureCredentials {
+  string subscription_id = 1;
+  string tenant_id = 2;
+  string client_id = 3;
+  string client_secret = 4;
+}
+
+message GcpCredentials {
+  string impersonation_token = 1;
```
michaeljguarino, line 22: I think this should be a service account json file

go/cloud-query/api/proto/cloudquery.proto
```diff
@@ -0,0 +1,74 @@
...
+  optional string query = 2;
+}
+
+message ExtractInput {
+  Connection connection = 1;
+}
+
+message QueryOutput {
+  repeated string columns = 1;
+  map<string, string> result = 2;
```
michaeljguarino, line 50: is there a way to have the value type be anything outside of string here?  Could be nice if we want to put real logic on these outputs

go/cloud-query/api/proto/cloudquery.proto
```diff
@@ -0,0 +1,74 @@
...
+}
+
+message SchemaOutput {
+  string table = 1;
+  repeated SchemaColumn columns = 2;
+}
+
+message ExtractOutput {
+  string type = 1;
+  map<string, string> result = 2;
```
michaeljguarino, line 65: same as above

go/cloud-query/api/proto/cloudquery.proto
```diff
@@ -0,0 +1,74 @@
...
+  string id = 3;
+  repeated string links = 4;
+}
+
+service CloudQuery {
+  rpc Query(QueryInput) returns (stream QueryOutput) {}
```
michaeljguarino, line 71: I wouldn't want Query or Schema to return streams if possible, just would make it easier to call

go/cloud-query/api/proto/cloudquery.proto
floreks: It's not really possible since protobuf does not support any type. It has to be a known type in the end. The closest thing here is string where complex values/objects are JSON encoded strings. That is what we are doing now. Column types can be found using schema api and it would be on the client to properly parse it.

go/cloud-query/api/proto/cloudquery.proto
maciaszczykm: See: https://hub.steampipe.io/plugins/turbot/gcp

go/cloud-query/api/proto/cloudquery.proto
maciaszczykm: Done.

michaeljguarino: Yea that's the ideal in my view (at least at the moment)

charts/cloud-query/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: cloud-query
```
michaeljguarino, line 2: It would honestly be easier to define this entirely within the console chart, but if you want it to be a separate chart, you'll need to add the same versioning logic we do for the `controller` chart in gh actions, since it'll be released with the same tagging all at once.

go/cloud-query/api/proto/cloudquery.proto
maciaszczykm: We missed the info that it should be possible to use service account JSON without a file. They use the file path in most of the places but it looks like it should accept JSON as well.

go/cloud-query/internal/config/aws.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	tmpl, err := template.New("connection").Parse(`
+		DROP SERVER IF EXISTS steampipe_{{ .ConnectionName }};
+		CREATE SERVER steampipe_{{ .ConnectionName }} FOREIGN DATA WRAPPER steampipe_postgres_aws OPTIONS (
+			config '
+				access_key="{{ .AccessKey }}"
+				secret_key="{{ .SecretKey }}"
+		');
+		IMPORT FOREIGN SCHEMA "{{ .ConnectionName }}" FROM SERVER steampipe_{{ .ConnectionName }} INTO "{{ .ConnectionName }}";
```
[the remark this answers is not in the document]
floreks: Connection name is not based on user input. We are generating unique UUID v6 and it is then used as the connection name, user and password. Every connection is a new user with new schema and permissions to only access its own schema.

michaeljguarino (approved):
I'm guessing this is fully ready to merge right?

PR #2353, opened 2025-06-11
Clusters Users Dump

JohnBlackwell:
This is for the MM V2 Workspace Report.

## Checklist

Plural Flow: console


michaeljguarino (commented):
this looks fine, is there a test for this though?  I'm pretty sure there's already an existing test that could be modified

michaeljguarino:
> Yeah I'll find the test. Not sure how comprehensive it was. What I've also been doing is pointing the console tag to the build sha and generating reports on plrl-aws cluster to make sure it's working.
> 
> Edit: I'm updating that test.

fwiw that makes your iteration time > 8-10m a pop, which will really slow things down.   It's usually better to just develop off the local tests as a result, which are nearly complete and fast

test/console_web/controllers/compliance_controller_test.exs
```diff
@@ -14,33 +14,23 @@ defmodule ConsoleWeb.ComplianceControllerTest do
 
       insert(:policy_binding, policy_id: cluster.write_policy_id, user_id: user2.id)
 
-      response = conn
+      clusters_content = Console.Compliance.Datasource.Clusters.stream()
```
michaeljguarino, line 17: this shouldn't be in a controller test, make it in a separate test, probably under `test/console/compliance` (the test files should match the code file structure).

test/console/compliance/clusters_test.exs
```diff
@@ -0,0 +1,24 @@
+defmodule Console.Compliance.Datasource.ClustersTest do
+  use Console.DataCase, async: true
+
+  test "it includes read and write users in the compliance report" do
+    user1 = insert(:user, email: "user1@example.com")
+    user2 = insert(:user, email: "user2@example.com")
+    group = insert(:group, name: "compliance-group")
+
+    cluster = insert(:cluster)
```
michaeljguarino, line 9: fwiw, a shorter way to set up the bindings here is:

```elixir
cluster = insert(:cluster, read_bindings: [%{user_id: user1.id}, %{group_id: group.id}], write_bindings: [%{user_id: user2.id}])
```

(most of the existing tests do that, and most associations can be instantiated similarly using the testdata lib)

test/support/factory.ex
```diff
@@ -1024,4 +1024,10 @@ defmodule Console.Factory do
   def bot(name) do
     insert(:user, bot_name: name, roles: %{admin: true})
   end
+
+  def policy_binding_factory do
```
michaeljguarino, line 1028: (that's why this factory was never defined)

PR #2356, opened 2025-06-12
PersistentVolume Evidence Module for AI

kinjalh:
Add PersistentVolume claims as AI evidence generated by the Cron job. The PVCs search CSI log drivers using provisioner and component name as filters.

The PVC module is also integrated into StatefulSet so that PersistentVolume data, if available, is added to the evidence for StatefulSet failures.

Added unit test that makes sure to check evidence string content.

## Test Plan
Unit test

## Checklist

Plural Flow: console

### DIFF
--- a/lib/console/ai/evidence/component/persistent_volume_claim.ex
+++ b/lib/console/ai/evidence/component/persistent_volume_claim.ex
@@ -8,21 +8,20 @@ defmodule Console.AI.Evidence.Component.PersistentVolumeClaim do
   alias Console.AI.Evidence.Context
 
   def hydrate(%CoreV1.PersistentVolumeClaim{} = pvc) do
-    messages = [
-      {:user, "the persistent volume claim #{component(pvc)} has a current state of:\n#{json_blob(encode(pvc))}"}
-    ] ++ pv_message(pvc)
-
-    messages
+    pv_message(pvc)
+    |> prepend({:user, "the persistent volume claim #{component(pvc)} has a current state of:\n#{json_blob(encode(pvc))}"})
     |> maybe_add_csi_logs(pvc, Repo.preload(get_cluster(), [:operational_layout]))
   end
   def hydrate(_), do: {:ok, []}
 
   defp pv_message(%CoreV1.PersistentVolumeClaim{spec: %{volume_name: name}}) when is_binary(name) do
-    with {:ok, pv} <- CoreV1.read_persistent_volume!(name) |> Kube.Utils.run() do
-      [
-        {:user, "the claim is bound to the persistent volume #{name}, which has a state of:\n#{json_blob(encode(pv))}"}
-      ]
-    else
+    CoreV1.read_persistent_volume!(name)
+    |> Kube.Utils.run()
+    |> case do
+      {:ok, pv} ->
+        [
+          {:user, "the claim is bound to the persistent volume #{name}, which has a state of:\n#{json_blob(encode(pv))}"}
+        ]
       _ -> []
     end
   end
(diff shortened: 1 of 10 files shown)


lib/console/ai/evidence/component/persistent_volume_claim.ex
```diff
@@ -0,0 +1,90 @@
...
+
+    with {:ok, query} <- {:ok, %{query | resource: cluster}},
+         {:ok, log_lines} <- Provider.query(query) do
+      # Format the logs similar to the original format
+      logs = log_lines
+             |> Enum.take(100)  # Equivalent to tail_lines: 100
+             |> Enum.map(& &1.log)
+             |> Enum.join("\n")
+
+      {:ok, logs, namespace}
```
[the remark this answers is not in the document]
kinjalh: sounds good i'll make the changes

lib/console/ai/evidence/component/persistent_volume_claim.ex
```diff
@@ -0,0 +1,63 @@
...
+    else
+      _ -> []
+    end
+  end
+  defp pv_message(_), do: []
+
+  defp maybe_add_csi_logs(history, %CoreV1.PersistentVolumeClaim{spec: %{storage_class_name: sc_name}}, %Cluster{
+    operational_layout: %OperationalLayout{namespaces: %OperationalLayout.Namespaces{} = ns}
+  } = cluster) when is_binary(sc_name) do
+    with {:ok, %StorageV1.StorageClass{provisioner: p}} <- StorageV1.read_storage_class!(sc_name) |> Kube.Utils.run(),
```
michaeljguarino, line 33: small stylistic tweak, but I think this might be more readable if you just pipe into case eg:

```elixir
StorageV1.read_storage_class!(sc_name)
|> Kube.Utils.run()
|> case do
  {:ok, %StorageV1.StorageClass{provisioner: p}} ->
     ...
  _ -> {:ok, history}
end
```

(generally with clauses are useful if you have > 1 sequential match or don't use the else clause)

lib/console/ai/evidence/component/persistent_volume_claim.ex
```diff
@@ -0,0 +1,63 @@
...
+         csi_namespace <- resolve_csi_namespace(p, ns) do
+
+      updated_history = history ++ [
```
michaeljguarino, line 36: can just do this as just one pipeline i think, eg:

```elixir
Enum.concat(history, [
  ...
])
|> Logs.with_logging(
  cluster, 
  force: true, 
  namespaces: [resolve_csi_namespace(p, ns)],
  q: p # I'm not sure this is the write query here, you probably just want to look for error logs
)
```

lib/console/ai/evidence/component/stateful_set.ex
```diff
@@ -1,9 +1,68 @@
...
   end
+
   def hydrate(_), do: {:ok, []}
+
+  defp pvc_messages(%AppsV1.StatefulSet{
+         metadata: %{name: sts_name, namespace: ns},
+         spec: %{replicas: replicas, volume_claim_templates: vcts}
+       })
+       when is_list(vcts) do
+    vcts
```
michaeljguarino, line 21: this should probably be written as a reduce into the Context struct, eg:

```elixir
Enum.reduce(vcts, Context.new(history), fn vct, acc ->
  Enum.reduce(0..(replicas - 1), acc, fn i, acc ->
     case fetch_pvc_evidence("#{vct.metadata.name}-#{sts_name}-#{i}") do
       {:ok, ctx} -> Context.merge(acc, ctx)
       _ -> acc
     end
  end)
end)
```

(I don't believe the Context.merge function has been implemented yet though, and needs to handle both the case where its given a context struct or just a message list)

lib/console/ai/evidence/component/stateful_set.ex
```diff
@@ -1,9 +1,68 @@
...
+        else
+          _ -> []
+        end
+      end)
+    end)
+  end
+
+  defp pvc_messages(_), do: []
+
+  defp evidence(observations) when is_list(observations) do
```
michaeljguarino, line 39: this should just be done from the above reduce and so isn't necessary

michaeljguarino:
Made a few changes for readability and to more elegantly leverage the context struct/only put the relevant stuff as evidence.  Figured this was a bit easier to sort of explain, but main things are:

* oftentimes case is cleaner than with if you need to pipe into it
* there were some funcs like Context.prompt/etc that makes it easier to compose the running history
* ideally it used a context as long as possible until the return value (at some point, hydrate functions should just be refactored to always return Context.t I think)

michaeljguarino (approved):

PR #2370, opened 2025-06-17
feat: add postgres datastore

zreigz:
## Checklist

Plural Flow: console


go/datastore/internal/client/postgres/client.go
```diff
@@ -0,0 +1,153 @@
...
+		return fmt.Errorf("secret %s does not contain key %s", credentials.Spec.PasswordSecretKeyRef.Name, credentials.Spec.PasswordSecretKeyRef.Key)
+	}
+
+	password := strings.ReplaceAll(string(key), "\n", "")
+
+	u := &url.URL{
+		Scheme: "postgres",
+		User:   url.UserPassword(credentials.Spec.Username, password),
+		Host:   fmt.Sprintf("%s:%d", credentials.Spec.Host, credentials.Spec.Port),
+		Path:   "postgres",
```
michaeljguarino, line 56: I don't think we should assume login to the postgres path, it could be in the credentials too (as database in the spec)

go/datastore/internal/client/postgres/client.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return fmt.Errorf("checking database existence: %w", err)
+	}
+
+	if exists {
+		return nil
+	}
+
+	// Create the database
+	_, err = db.Exec(fmt.Sprintf(`CREATE DATABASE "%s"`, dbName))
```
michaeljguarino, line 147: Isn't there a `CREATE DATABASE IF NOT EXISTS ...` you can use?

PR #2406, opened 2025-06-27
chore(cloud-query): merge cloud-query chart with the console chart

floreks:
```
TBD
```

## Checklist

Plural Flow: console


charts/console/values.yaml
```diff
@@ -775,3 +775,90 @@ dashboard:
       # ServiceMonitor connection tlsConfig. Defaults to {insecureSkipVerify:true}.
       tlsConfig:
         insecureSkipVerify: true
+
+
+cloudQuery:
+  replicaCount: 1
```
michaeljguarino, line 781: we should add an enabled flag here, which would drop at minimum the deployment + service resources

PR #2417, opened 2025-07-01
Implement NamespaceManagement CRD

michaeljguarino:
First pass at namespace management implementation

## Test Plan
local

## Checklist

Plural Flow: console


zreigz:
I did some improvements:
 - Register controller
 - added a generic function `getResource` to check the sentinel (now it can work with any type)
 - used `Sentinel` type instead of `corev1.ObjectReference` to make sure the APIVersion and Kind are set
 - added unit test to confirm the code logic

zreigz (approved):

michaeljguarino:
@zreigz did you test this against a running cluster too?

PR #2423, opened 2025-07-03
feat: enable gRPC extract streaming for aws provider

floreks:
- Added AWS extractor with an initial support for resources:
  - `aws_vpc`
  - `aws_vpc_subnet`
  - `aws_vpc_peering_connection`
  - `aws_eks_addon`
  - `aws_eks_cluster`
  - `aws_eks_node_group`
  - `aws_ec2_instance`
  - `aws_s3_bucket`
- Added a way to enforce ordered resource extraction
- Added support for a lookup table that is filled incrementally while extracting resources and is passed to build a proper links array.
- Fixed `Schema` query statement issue

Output is streamed resource by resource, row by row.

## Test Plan

Locally

## Checklist

Plural Flow: console


go/cloud-query/internal/extractor/extractor_default.go
```diff
@@ -0,0 +1,80 @@
...
+type DefaultExtractor struct {
+	sink      Sink
+	resources map[string]ResourceFactory
+}
+
+// Extract extracts data from the provided connection and sends it to the sink.
+func (in DefaultExtractor) Extract(conn connection.Connection) error {
+	klog.V(log.LogLevelDebug).InfoS("starting extraction")
+
+	for _, entry := range lo.Entries(in.Resources()) {
```
michaeljguarino, line 29: why do you need to wrap this in `lo.Entries`?

go/cloud-query/internal/extractor/extractor_default.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to query table '%s': %w", table, err)
+		}
+
+		klog.V(log.LogLevelDebug).InfoS("extracting table", "table", table, "columns", len(columns), "rows", len(rows))
+		for _, row := range rows {
+			klog.V(log.LogLevelTrace).InfoS("processing row", "table", table, "row", row)
+			rowJson, err := common.ToRowJSON(columns, row)
+			if err != nil {
+				return fmt.Errorf("failed to convert row to JSON: %w", err)
```
michaeljguarino, line 42: I'm wondering if instead of breaking we should continue on these

michaeljguarino:
You're also going to eventually need to add support for reducing a full lookup table of ids and communicating it back to the resources here, but that could be deferred too.

go/cloud-query/internal/extractor/extractor_default.go
floreks: It would definitely be more resilient to random errors. In that case it makes sense to continue on error.

go/cloud-query/internal/extractor/extractor_default.go
```diff
@@ -0,0 +1,89 @@
...
+				klog.V(log.LogLevelVerbose).ErrorS(err, "failed to convert row to JSON", "table", table, "row", row)
+				continue
+			}
+
+			itable, err := factory(rowJson)
+			if err != nil {
+				return fmt.Errorf("could not create table interface: %w", err)
+			}
+
+			in.lookup[itable.ShortID()] = itable.ID()
```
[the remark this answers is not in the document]
michaeljguarino: I think in aws especially often times resources will have different identifiers embedded in them, eg one might have an arn, one might have a subnet id, etc.  So you need to be able to keep around all the id variants to lookup appropriately.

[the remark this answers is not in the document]
floreks: I'd leave that for now. If we'll hit a case where lookup will need to be extended, we can do it then.

michaeljguarino (approved):

PR #2453, opened 2025-07-10
feat(cloud-query): add PostgreSQL sidecar container and update build process

floreks:
- Added separate dockerfile to prepare dedicated postgres sidecar container for cloud query
- Added a simple `/healthz` endpoint to the cloud query server to do a db connection check. It is used as `readinessProbe`.
- Updated GH actions to publish a new `cloud-query-db` image.
- Reconfigured cloud query chart to use a db sidecar container. It does not expose any ports and db is only accessible internally inside the pod network. `cloud-query` container uses localhost connection to access it.
- Updated `GcpCredentials` proto to include `project` as it is required by the gcp extension if gcloud CLI is not configured.

NOTE: Since by default both cloud-query and cloud-query-db images are using chart version, it will require a console release to trigger CD pipeline and publish required images.

## Test Plan

Did a `helm package` and tested whole cloud query chart setup on `plrl-dev-aws` env w/ `kubectl port-forward` to a service. Tested Query/Schema/Extract (AWS) API.

## Checklist

Plural Flow: console

### DIFF
--- a/charts/console-rapid/Chart.yaml
+++ b/charts/console-rapid/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: console-rapid
 description: rapid channel chart for the plural console (used for testing)
 appVersion: 0.11.28
-version: 0.3.132
+version: 0.3.133
 dependencies:
   - name: kas
     version: 0.3.0


go/cloud-query/db.Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+
+WORKDIR /workspace
+
+COPY hack/ hack/
+
+ARG TARGETOS
+ARG TARGETARCH
+
+# Set gcloud CLI download URL based on architecture
+RUN case ${TARGETARCH} in \
```
michaeljguarino, line 20: why do you need to download gcloud?

floreks:
### Vulnerability scan of the db image
The biggest issues are:
- the outdated postgres 
- postgres fdw extensions for aws/gcp/azure provider

Unfortunately, we cannot update those since we are using latest extensions and they are only released for pg 15 currently. `gcloud` binary itself reports no vulnerabilities and compared to other issues should be safe to be kept inside. Let me know if you'd still want to built a dedicated image without the gcloud binary.
image

michaeljguarino:
It's possible gcloud has gotten better, let's just stick to that for now.

michaeljguarino:
also can you bump the chart version for `console-rapid` (not `console`, that will distribute to all users).  That way i can pick up the changes and test easily.

floreks:
> also can you bump the chart version for `console-rapid` (not `console`, that will distribute to all users). That way i can pick up the changes and test easily.

sure, done

PR #2464, opened 2025-07-15
feat(stacks): add Terraform configuration options and tags management

floreks:
- Updated stacks overview UI as per [Figma](https://www.figma.com/design/Kn0s2plNxcvpPrlau72z1v/💻-Console?node-id=12688-130888&t=HZBn4dGIpnTtIsgc-0)

## Test Plan

Locally

## Checklist

Plural Flow: console

### DIFF
--- a/assets/src/components/stacks/overview/StackConfiguration.tsx
+++ b/assets/src/components/stacks/overview/StackConfiguration.tsx
@@ -4,7 +4,7 @@ import { useState } from 'react'
 import { useOutletContext, useParams } from 'react-router-dom'
 import { useTheme } from 'styled-components'
 
-import { useUpdateStackMutation } from '../../../generated/graphql'
+import { StackType, useUpdateStackMutation } from '../../../generated/graphql'
 import { GqlError } from '../../utils/Alert'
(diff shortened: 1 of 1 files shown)


assets/src/components/stacks/overview/StackConfiguration.tsx
```diff
@@ -1,25 +1,31 @@
...
 import { StackOutletContextT } from '../Stacks'
-import { OverlineH1 } from '../../utils/typography/Text'
 
 export default function StackConfiguration() {
   const theme = useTheme()
   const { stackId = '' } = useParams()
   const { stack, refetch } = useOutletContext() as StackOutletContextT
   const [image, setImage] = useState(stack.configuration.image)
   const [version, setVersion] = useState(stack.configuration.version)
+  const [parallelism, setParallelism] = useState(
```
michaeljguarino, line 19: these fields should only be set if its a terraform stack, i think nick's mocks did a poor job of conveying that though

floreks: ahh... that makes sense. I will hide them if it's not tf.

michaeljguarino (approved):

PR #2465, opened 2025-07-15
fix(cloud-query): icrease default db connection TTL and improve cleanup logic

floreks:
- Fix connection cleanup logic
- Lock remove method to avoid a situation where cleanup timer collides with get and remove.
- Improve error handling and error logging
- Increase db connection TTL to 3h by default

## Test Plan

Locally

## Checklist

Plural Flow: console

### DIFF
--- a/charts/console-rapid/Chart.yaml
+++ b/charts/console-rapid/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: console-rapid
 description: rapid channel chart for the plural console (used for testing)
 appVersion: 0.11.29
-version: 0.3.134
+version: 0.3.135
 dependencies:
   - name: kas
     version: 0.3.0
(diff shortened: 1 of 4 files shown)


charts/console-rapid/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: console-rapid
 description: rapid channel chart for the plural console (used for testing)
 appVersion: 0.11.29
-version: 0.3.134
+version: 0.3.135
```
michaeljguarino, line 5: don't need to do this bump since no chart structure change

go/cloud-query/cmd/args/args.go
```diff
@@ -125,7 +125,7 @@ func DatabaseConnectionTTL() time.Duration {
 		return defaultDatabaseConnectionTTL
 	}
 
-	if *argDatabaseConnectionTTL < 1*time.Minute {
+	if *argDatabaseConnectionTTL < 10*time.Minute {
```
michaeljguarino, line 128: this should be way longer, like 2-5h, we'll continuously extract from the connection once an hour, so if its longer than that it likely no longer exists

PR #2468, opened 2025-07-16
feat(personas): add AI persona fields and integrate with PR creation logic

floreks:
- Make `Create PR` in the ai suggested fix respect persona ai configuration to show/hide the button.
- Add generic `hasAccess` method to check persona access

Some additional questions:
~~1. Should null/undefined value default to show or hide the button?~~
~~2. There is also a `Create PR` button as part of the chat pr call content. I am not fully sure how to trigger it to show, but should same logic also be applied there?~~

## Test Plan

Locally

## Checklist

Plural Flow: console


assets/src/components/ai/chatbot/AISuggestFix.tsx
```diff
@@ -121,15 +123,15 @@ function FixPr({
         >
           View PR
         </Button>
-      ) : (
+      ) : personaConfiguration?.ai?.pr !== false ? (
```
michaeljguarino, line 126: we should also respect this if the top level `all` flag is set to true.

jsladerman:
we actually removed the create pr button from chat recently in favor of the new agent setup

assets/src/components/ai/chatbot/AISuggestFix.tsx
floreks: Done. I have added generic typesafe method to check that.

PR #2484, opened 2025-07-22
WIP: new chat UI

jsladerman:


michaeljguarino:
Took a look at current progress, here's my quick hits:

1. Action panel should auto-expand the moment there's something in it
2. Service/stacks seem like they're probably in progress but they definitely need a link out (probably also true of pull requests).  The idea is they're supposed to be CTAs
3. copilot should default closed
4. I saw a weird bug when going through provisioning where it didn't show the generated messages after approving the implementation plan.  It's possible that whole experience is suboptimal though and should be reconsidered.

maciaszczykm:
> copilot should default closed

@michaeljguarino I'll use persisted state for that as we have discussed it during one of the daily meetings. Should it then default to closed as well or maybe we should open it for discoverability?

floreks:
Fixed point 5 too

michaeljguarino:
horizontal space issues:

* cd clusters tabs have no horizontal scroll
* stacks tabs have no horizontal scrolll
*  vulnerability reports selects get weirdly scrunched

michaeljguarino:
one last piece of feedback for copilot ui:

stacks/services should default expanded in the action panel

Otherwise looks good to merge

PR #2492, opened 2025-07-25
add gzip support for  gql responses

zreigz:
I've added Gzip support for GraphQL responses.

## Checklist

Plural Flow: console


michaeljguarino:
It looks like phoenix only has out of the box support for response compression not request compression, due to some security concerns.  We might need to figure out how to support both server-side first

michaeljguarino (commented):
could we reconfigure this to just gzip responses and not requests?

PR #2505, opened 2025-07-29
feat(controller): add support for FederatedCredential CRD

floreks:
- Added support for new `FederatedCredential` CRD

## Test Plan

Locally against dev env

## Checklist

Plural Flow: console


go/controller/api/v1alpha1/federatedcredential_types.go
```diff
@@ -0,0 +1,86 @@
...
+	Issuer string `json:"issuer"`
+
+	// Scopes are the scopes that the credential will request from the identity provider.
+	// +kubebuilder:validation:Optional
+	Scopes []string `json:"scopes,omitempty"`
+
+	// ClaimsLike is a JSON expression that matches the claims in the token.
+	// TODO: describe the syntax of this expression.
+	// +kubebuilder:validation:Optional
+	ClaimsLike *string `json:"claimsLike,omitempty"`
```
michaeljguarino, line 80: We should make this a raw extension object so users don't have to manually json format (then you json encode it w/in the operator)

floreks: makes sense, done

michaeljguarino (approved):

michaeljguarino:
> > the actual way the claims like value would work would be:
> > ```yaml
> > claimsLike:
> >    sub: "repo:floreks/chartmuseum:ref:refs/heads/.*"
> > ```
> 
> oh yeah, it's just a raw object extension now. I have just taken some example from microsoft federated credential examples. I have updated test yaml in my original comment. If there is an actual schema behind this, I can also update code doc to reflect that.


The schema is just map[string]string basically on the backend, with a validation that all the value strings are valid regexes

floreks:
Done. Slightly updated `claimsLike` code docs.

PR #2506, opened 2025-07-29
refactor(scmconnection): use custom annotation to reconcile on secret ref changes

floreks:
This change ensures that secrets referenced by the scm connection are triggering a scm reconcile.

- Added a new common `deployments.plural.sh/owned-by` annotation handling
- Added a common `OwnedByEventHandler` that can filter `owned-by` annotation based on the provided owner group/kind
- If exist, old owner references will be removed from the referenced secret
- If annotation does not match expected format, it updates it to match `<group>/<kind>/<namespace>/<name>` format.

#### Test YAML
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: github-token
  namespace: default
type: Opaque
stringData:
  token: "YOUR_GITHUB_TOKEN"
---
apiVersion: deployments.plural.sh/v1alpha1
kind: ScmConnection
metadata:
  name: github
spec:
  name: github
  type: GITHUB
  username: plural
  tokenSecretRef:
    name: github-token
    namespace: default
```

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: github-token
  namespace: default
  ownerReferences:
    - apiVersion: deployments.plural.sh/v1alpha1
      kind: ScmConnection
      name: github
      uid: "SCM_UID"
type: Opaque
stringData:
  token: "YOUR_GITHUB_TOKEN"
---
apiVersion: deployments.plural.sh/v1alpha1
kind: ScmConnection
metadata:
  name: github
spec:
  name: github
  type: GITHUB
  username: plural
  tokenSecretRef:
    name: github-token
    namespace: default
```

## Checklist

Plural Flow: console


go/controller/internal/controller/scmconnection_controller.go
```diff
@@ -100,11 +100,6 @@ func (r *ScmConnectionReconciler) Reconcile(ctx context.Context, req reconcile.R
 	// Mark resource as managed by this operator.
 	utils.MarkCondition(scm.SetCondition, v1alpha1.ReadonlyConditionType, v1.ConditionFalse, v1alpha1.ReadonlyConditionReason, "")
 
-	err = r.tryAddControllerRef(ctx, scm)
```
michaeljguarino, line 103: if there's no owner ref we still need a mechanism to propagate updates from the secret to the scm connection.  Usually this is done via a watches + annotation in other controllers.

floreks: ahh, ok. I thought you wanted to keep it completely separate. I will update it.

go/controller/internal/controller/scmconnection_controller.go
```diff
@@ -100,8 +100,7 @@ func (r *ScmConnectionReconciler) Reconcile(ctx context.Context, req reconcile.R
...
-	if err != nil {
+	if err := TryAddOwnedByAnnotation(ctx, r.Client, scm); err != nil {
```
michaeljguarino, line 103: we should also try to remove old controller refs if possible (might not be worth)

PR #2560, opened 2025-08-12
feat: Add mgmt column to global services table

maciaszczykm:


jsladerman (approved):

assets/src/components/cd/globalServices/GlobalServicesColumns.tsx
```diff
@@ -86,6 +91,27 @@ export const ColProject = columnHelper.accessor(
   }
 )
 
+export const ColMgmt = columnHelper.accessor(({ node }) => node?.mgmt, {
+  id: 'mgmt',
+  header: 'Mgmt',
+  cell: ({ getValue }) =>
+    getValue() ? (
```
michaeljguarino, line 98: It will be included if the value is undefind as well (only excluded if explicitly false)

maciaszczykm: Fixed the check.

PR #2569, opened 2025-08-14
use registry cache

kinjalh:
## Test Plan
tested by running action, seems to make progress towards not hitting rate limit.
Timing increase is somewhat marginal (only ~2 min faster, from 8 to 6 mins)
For now doesn't seem like there is a way to make build times very much faster, might revisit if any ideas come up. Still probably worth making this change to avoid having to re-run PR actions due to rate limit failures.

## Checklist

Plural Flow: console

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -71,8 +71,8 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           platforms: linux/amd64
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          cache-from: type=registry,ref=ghcr.io/pluralsh/console:buildcache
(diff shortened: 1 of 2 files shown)


michaeljguarino (approved):
Could we also try flipping the runs

.github/workflows/test.yaml
```diff
@@ -71,8 +71,8 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           platforms: linux/amd64
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          cache-from: type=registry,ref=ghcr.io/pluralsh/console:buildcache
```
michaeljguarino, line 74: I think it's probably wise to use a slightly different repo name (pluralsh/console-cache) for this

PR #2571, opened 2025-08-14
Verify github webhook signature

JohnBlackwell:
## Checklist

Plural Flow: console

### DIFF
--- a/go/controller/pkg/webhooks/verify.go
+++ b/go/controller/pkg/webhooks/verify.go
@@ -1,0 +1,57 @@
+package webhooks
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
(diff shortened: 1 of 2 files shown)


go/sign/sign.go
```diff
@@ -0,0 +1,72 @@
...
+	Body   map[string]interface{} `json:"body"`
+}
+
+func signBody(secret, body []byte) string {
+	computed := hmac.New(sha256.New, secret)
+	computed.Write(body)
+	return "sha256=" + hex.EncodeToString(computed.Sum(nil))
+}
+
+func sendWebhookRequest(url string, secret []byte, body []byte) error {
```
michaeljguarino, line 28: we should have this function return the response body from the downstream http endpoint (i'm guessing there's logs that would make that useful to have)

PR #2586, opened 2025-08-18
Allow for ping interval customization in cluster health

michaeljguarino:
Currently it will always declare unhealthy if no ping in 20m, this will allow it to work with any customized ping from an AgentConfiguration

Fixes PROD-3879

## Test Plan
n/a

## Checklist

Plural Flow: console


go/controller/internal/controller/common.go
```diff
@@ -48,6 +51,10 @@ var (
 	waitForResources = ctrl.Result{RequeueAfter: requeueWaitForResources}
```
maciaszczykm, line 51: Should we also add jitter when it's waiting for resources? It's used in a lot of cases as `handleRequeue` uses it.

maciaszczykm (approved):
Left one question.

go/controller/internal/controller/common.go
michaeljguarino: yea we definitely should i just didn't get to it.  I'd add another ticket to handle it

PR #2592, opened 2025-08-20
Implement key rotation on ai proxy

JohnBlackwell:
## Checklist

Plural Flow: console

### DIFF
--- a/go/ai-proxy/args/args.go
+++ b/go/ai-proxy/args/args.go
@@ -27,7 +27,7 @@ const (
 var (
 	argProvider               = pflag.String("provider", defaultProvider.String(), "Provider name. Must be one of: ollama, openai, vertex. Defaults to 'ollama' type API.")
 	argProviderHost           = pflag.String("provider-host", "", "Provider host address to access the API i.e. https://api.openai.com")
-	argProviderToken          = pflag.String("provider-token", helpers.GetPluralEnv(envProviderToken, ""), "Provider token used to connect to the API if needed. Can be overridden via PLRL_PROVIDER_TOKEN env var.")
+	argProviderTokens         = pflag.StringSlice("provider-tokens", []string{helpers.GetPluralEnv(envProviderToken, "")}, "Provider tokens used to connect to the API if needed. Can be overridden via PLRL_PROVIDER_TOKEN env var.")
 	argProviderServiceAccount = pflag.String("provider-service-account", helpers.GetPluralEnv(envProviderServiceAccount, ""), "Provider service account file used to connect to the API if needed. Can be overridden via PLRL_PROVIDER_SERVICE_ACCOUNT env var.")
 	argsProviderAWSRegion     = pflag.String("provider-aws-region", helpers.GetPluralEnv(envProviderAWSRegion, ""), "Provider AWS region used to connect to BedRock API.")
 	argPort                   = pflag.Int("port", defaultPort, "The port to listen on. Defaults to port 8000.")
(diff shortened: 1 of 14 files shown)


go/ai-proxy/proxy/bedrock/bedrock.go
```diff
@@ -30,7 +30,12 @@ type BedrockProxy struct {
 	bedrockClient *bedrockruntime.Client
 }
 
-func NewBedrockProxy(region string) (api.OpenAIProxy, error) {
+func NewBedrockProxy(credentials []string) (api.OpenAIProxy, error) {
+	var region string
```
michaeljguarino, line 34: why is region coming from the credentials?

go/ai-proxy/proxy/openai/openai.go
```diff
@@ -14,8 +17,63 @@ import (
...
+
+func (rr *RoundRobinTokenRotator) AddToken(token string) {
+	rr.mu.Lock()
+	defer rr.mu.Unlock()
+
+	rr.tokens = append(rr.tokens, token)
+	rr.index.Store(0) // Reset index when modifying tokens
+}
+
+func (rr *RoundRobinTokenRotator) RemoveToken(token string) {
```
michaeljguarino, line 65: why is there a remove token function? You should just have this cycle and never mutate the underlying list

go/ai-proxy/proxy/openai/openai.go
```diff
@@ -14,8 +17,63 @@ import (
...
+func (rr *RoundRobinTokenRotator) GetNextToken() string {
+	rr.mu.RLock()
+	defer rr.mu.RUnlock()
+
+	if len(rr.tokens) == 0 {
+		return ""
+	}
+
+	currentIndex := rr.index.Add(1) - 1
+	return rr.tokens[currentIndex%uint32(len(rr.tokens))]
```
michaeljguarino, line 54: isn't mod 0-basis?  the -1 should be irrelevant

go/ai-proxy/proxy/openai/openai.go
michaeljguarino: We'd reconfigure the entire process with a new key, you can assume they're always valid if used

go/ai-proxy/proxy/bedrock/bedrock.go
JohnBlackwell: https://github.com/pluralsh/console/blob/master/go/ai-proxy/proxy/proxy.go#L33

go/ai-proxy/args/args.go
```diff
@@ -76,24 +76,24 @@ func ProviderHost() string {
 	return *argProviderHost
 }
 
-func ProviderCredentials() string {
-	if len(*argProviderToken) > 0 && Provider() == api.ProviderOpenAI {
-		return *argProviderToken
+func ProviderCredentials() []string {
```
michaeljguarino, line 79: this function seems just really ill-designed. We shouldn't have it return totally dissimilar things like an openai key, region name and service account

go/ai-proxy/args/args.go
```diff
@@ -76,27 +76,26 @@ func ProviderHost() string {
 	return *argProviderHost
 }
 
-func ProviderCredentials() []string {
-	if len(*argProviderTokens) > 0 && Provider() == api.ProviderOpenAI {
-		return *argProviderTokens
-	}
-
+func ProviderServiceAccount() string {
 	if len(*argProviderServiceAccount) > 0 && Provider() == api.ProviderVertex {
```
michaeljguarino, line 80: won't this nil pointer?

go/ai-proxy/args/args.go
```diff
@@ -76,27 +76,26 @@ func ProviderHost() string {
...
+	return ""
+}
 
-	if Provider() == api.ProviderBedrock {
-		if len(*argsProviderAWSRegion) > 0 {
-			return []string{*argsProviderAWSRegion}
-		}
-		return []string{}
+func ProviderAwsRegion() string {
+	if Provider() == api.ProviderBedrock && len(*argsProviderAWSRegion) > 0 {
```
michaeljguarino, line 87: same this is almost certainly a nil pointer risk

go/ai-proxy/args/args.go
```diff
@@ -76,27 +76,26 @@ func ProviderHost() string {
...
+		return *argsProviderAWSRegion
 	}
+	return ""
+}
 
-	if Provider() == defaultProvider {
-		return []string{}
+func ProviderTokens() []string {
+	if len(*argProviderTokens) > 0 && Provider() == api.ProviderOpenAI {
```
michaeljguarino, line 94: nil pointer

go/ai-proxy/internal/helpers/token_rotator.go
```diff
@@ -0,0 +1,38 @@
...
+	tokenCpy := make([]string, len(tokens))
+	copy(tokenCpy, tokens)
+
+	return &RoundRobinTokenRotator{
+		Tokens: tokenCpy,
+	}
+}
+
+func (rr *RoundRobinTokenRotator) GetNextToken() string {
+	rr.mu.RLock()
```
michaeljguarino, line 28: do you need to mutex if the add is already atomic?

go/ai-proxy/args/args.go
JohnBlackwell: It shouldn't. It defaults to an empty string if nothing is provided https://github.com/pluralsh/console/blob/15ffa3b932d8b3715ef94567d9b036dfdda22ed1/go/ai-proxy/args/args.go#L31

go/ai-proxy/args/args.go
JohnBlackwell: https://github.com/pluralsh/console/blob/15ffa3b932d8b3715ef94567d9b036dfdda22ed1/go/ai-proxy/args/args.go#L32 

same check here

go/ai-proxy/args/args.go
michaeljguarino: We should still fix the codepaths, it's entirely possible a future developer doesn't know to guard it upstream of that callsite.

go/ai-proxy/args/args.go
michaeljguarino: same as above, it's entirely possible a future developer doesn't know to guard it upstream of that callsite.

michaeljguarino (approved):

PR #2618, opened 2025-08-26
feat: Build extended support matrices for aks/gke/eks

JohnBlackwell:
## Checklist

Plural Flow: console


static/extended/aks.yaml
```diff
@@ -0,0 +1,26 @@
+- version: 1.33
+  extended: true
```
michaeljguarino, line 2: I think the boolean here is upside down

.github/workflows/compatability-matrices.yaml
```diff
@@ -47,3 +48,8 @@ jobs:
         run: |
           aws s3 cp ./static/versions.yml s3://plural-www-assets/matrices/compatability/static/versions.yml
         if: hashFiles('static/versions.yml') != ''
+
+      - name: Upload static/extended directory to S3
```
michaeljguarino, line 52: you also need to call your new script in this github action somewhere right?

static/extended/aks.yaml
JohnBlackwell: Why? https://endoflife.date/azure-kubernetes-service

michaeljguarino: The recent versions are not on extended support, the old versions are.  It's inverted

static/extended/gke.yaml
```diff
@@ -0,0 +1,34 @@
+- version: 1.33
+  extended: false
+- version: 1.32
```
michaeljguarino, line 3: these versions should be strings not floats as well (can see it very clearly with `version: 1.3` instead of `version: 1.30` below

michaeljguarino (approved):

michaeljguarino (approved):
The output is fine, but might want to consider using their api

utils/extended/main.py
```diff
@@ -0,0 +1,57 @@
...
+import re
+
+SERVICES = [
+    "amazon-eks",
+    "azure-kubernetes-service",
+    "google-kubernetes-engine"
+]
+
+def fetch_extended_versions(service):
+    url = f"https://endoflife.date/{service}"
```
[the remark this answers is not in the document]
JohnBlackwell: i'll refactor. will be quick

michaeljguarino (approved):

PR #2645, opened 2025-09-05
feat: add image information to ComponentAttributes GraphQL schema

JohnBlackwell:
## Checklist

Plural Flow: console


michaeljguarino:
There is already a metadata column on the service that's meant to house these, i don't know if its exposed via graphql, but this definitely won't work.

michaeljguarino (changes requested):
this should be closed in favor of above

michaeljguarino:
(just check it is exposed in the updateServiceComponents mutation, there's no schema work needed here)

JohnBlackwell:
Closing this per comments above

PR #2646, opened 2025-09-05
add azure devops to scm configuration form

jsladerman:
consolidates/simplifies the logic for this form which has been finicky in the past
also adds icon frame links to the agent sessions table if the session has a pr, stack, or service (closes PROD-3875), and bumps Dockerfile node vsn from 16 to 20

Plural Flow: console


assets/src/components/self-service/pr/scm/CreateScmConnection.tsx
```diff
@@ -142,18 +121,31 @@ export function CreateScmConnection({
   )
 }
 
+// all forms need a name
+// gh auth only needs gh info (no token)
+// all other forms require a token
+// azure devops also requires azure info
+export const isValidScmForm = (
+  formState: ScmConnectionAttributes,
+  requireToken = true
```
michaeljguarino, line 130: is this ever set to false?

michaeljguarino: kk that works

michaeljguarino (approved):

PR #2664, opened 2025-09-10
add logs bucketing for elastic and opensearch

kinjalh:
Add log bucketing for elastic and opensearch.
To my knowledge Victoria does not offer this feature so leave it stubbed for that provider.

## Test Plan
unit tests

## Checklist

Plural Flow: console

### DIFF
--- a/lib/console/logs/provider/elastic.ex
+++ b/lib/console/logs/provider/elastic.ex
@@ -4,7 +4,7 @@ defmodule Console.Logs.Provider.Elastic do
   """
   @behaviour Console.Logs.Provider
   alias Console.Schema.{Cluster, Service, DeploymentSettings.Elastic}
-  alias Console.Logs.{Query, Line, Time}
+  alias Console.Logs.{Query, Line, Time, AggregationBucket}
 
   @type t :: %__MODULE__{}
   @headers [{"Content-Type", "application/json"}]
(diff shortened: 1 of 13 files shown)


lib/console/logs/provider/elastic.ex
```diff
@@ -50,6 +59,31 @@ defmodule Console.Logs.Provider.Elastic do
   end
   defp format_hits(_), do: []
 
+  defp format_aggregation_response(%Snap.SearchResponse{aggregations: aggs}) do
+    case aggs do
+      %{"logs_over_time" => %Snap.Aggregation{buckets: buckets}} ->
+        Enum.map(buckets, fn bucket ->
+          %{
```
michaeljguarino, line 66: could you make a struct for this as well?  Will make it more straightforward to understand the output typing

lib/console/logs/provider/elastic.ex
```diff
@@ -126,4 +160,28 @@ defmodule Console.Logs.Provider.Elastic do
 
   defp sort(%Query{time: %Time{reverse: true}}), do: [%{"@timestamp": %{order: "asc"}}]
   defp sort(_), do: [%{"@timestamp": %{order: "desc"}}]
+
+  defp build_aggregation_query(%Query{query: str} = q) do
+    res = %{
```
michaeljguarino, line 165: (also don't need to assign this variable, can just return the map directly

PR #2682, opened 2025-09-15
Add more agent runtime types

michaeljguarino:
Main one is a custom type, but also gemini seems like one we could support at some time

## Test Plan
n/a

## Checklist

Plural Flow: console


maciaszczykm (approved):
cc @zreigz

PR #2693, opened 2025-09-20
feat: updated opentelemetry compatibility and added a scraper

smyja:
https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/compatibility.md#compatibility-matrix

## Checklist

Plural Flow: console


michaeljguarino (commented):
This table is actually incorrectly named, it should be `opentelemetry-operator` since that's the name declared by the helm chart

PR #2695, opened 2025-09-22
feat: Change health score and cluster upgrades color when agent is unhealthy

maciaszczykm:
## Test Plan
Zrzut ekranu 2025-09-22 o 17 04 08

## Checklist

Plural Flow: console


michaeljguarino:
Let's not grey out the upgradeability nub.  I suppose it could also be considered invalid but its a bit different

michaeljguarino (approved):

PR #2726, opened 2025-10-02
Add images to services csv

JohnBlackwell:
## Checklist

Plural Flow: console


michaeljguarino (approved):
this should ultimately be inferrable from the vulnerability report dump already in place, it will also have cve info in it

michaeljguarino:
you probably need to solve for encoding the list as well, which a csv might struggle with

PR #2730, opened 2025-10-03
fix: Fix OCI Auth sidecar CVE

maciaszczykm:
Fixes https://github.com/pluralsh/console/security/dependabot/204.


michaeljguarino (approved):
if ths is getting built and tagged on each release, we should make sure the chart is always using the latest tag too to make sure it picks up any changes.  Currently it doesn't

PR #3421, opened 2026-04-13
feat(cloud-query): add Azure integration with support for metrics and logs queries

floreks:
This pull request adds support for Azure Monitor as a configurable tool in the workbench, enabling users to connect to Azure and query metrics and logs from Azure resources. The changes span the frontend, backend Go models, GraphQL schema, and documentation to fully integrate Azure as a first-class observability integration.

The most important changes are:

**Frontend integration and configuration:**
* Added `WorkbenchToolType.Azure` throughout the frontend, including configuration forms, tool type lists, labels, categories, and icon mapping in `WorkbenchToolForm.tsx` and `workbenchToolsUtils.tsx`. 
* Implemented `AzureFormFields` in `WorkbenchToolFormFields.tsx` to collect Azure credentials and resource information from the user.

**Backend Go model and API updates:**
* Added new Go types for Azure connections and configuration attributes, and updated the tool type enum and validation logic to include Azure in `models_gen.go`.

**GraphQL and gRPC/protobuf support:**
* Extended the GraphQL schema to expose Azure connection fields in `workbench.graphql`.
* Updated `toolquery.proto` to add `AzureConnection`, support Azure as a `ToolConnection`, and introduced Azure-specific options for metrics queries. 

**Documentation:**
* Updated API reference (`docs/api-reference.md`) and README to document Azure Monitor support, required credentials, query options, and usage details.

These changes collectively enable users to configure and use Azure Monitor as an observability tool within the workbench, with full support across the stack.

## Test Plan

Locally and on dev env

## Checklist

Plural Flow: console


floreks:
@greptileai there are no compilation failures. `new` is used correctly according to new Go 1.26 syntax. Remember this and review again.

floreks:
@greptileai review

lib/console/graphql/deployments/workbench.ex
```diff
@@ -510,6 +467,13 @@ defmodule Console.GraphQl.Deployments.Workbench do
     field :role_session_name, :string, description: "assume-role session name"
   end
 
+  object :workbench_tool_azure_connection do
+    field :subscription_id, :string, description: "azure subscription id"
+    field :tenant_id,       :string, description: "azure tenant id"
+    field :client_id,       :string, description: "azure client id"
+    field :resource_id,     :string, description: "azure resource id for metrics/logs queries"
```
michaeljguarino: i don't think this should be on the schema, instead it would be provided by the ai and passed dynamically in the tool call

michaeljguarino: (unless its relatively fixed, in which case we should make the description here a bit more clear that it's for the azure metrics workspace id or whatever

floreks: so it would be a part of azure options in logs/metrics query input?

floreks: it's definitely dynamic, not something that can just be set like credentials

michaeljguarino: Ok yea, in that case it's part of the query input and we'll have to make sure the ai can fill in the right value

floreks:
@greptileai review

floreks:
@greptileai Remember about go 1.26 language changes
> Changes to the language[¶](https://go.dev/doc/go1.26#language)
The built-in new function, which creates a new variable, now allows its operand to be an expression, specifying the initial value of the variable.

>
>This feature is particularly useful when working with serialization packages such as encoding/json or protocol buffers that use a pointer to represent an optional value, as it enables an optional field to be populated in a simple expression

Review again.

floreks:
@greptileai review

lib/console/ai/tools/workbench/observability/metrics.ex
```diff
@@ -3,14 +3,26 @@ defmodule Console.AI.Tools.Workbench.Observability.Metrics do
...
-  alias Toolquery.{MetricsQueryInput, MetricsQueryOutput, MetricPoint}
+  alias Toolquery.{MetricsQueryInput, MetricsQueryOutput, MetricsOptions, AzureMetricsOptions, MetricPoint}
   alias Console.AI.Workbench.Conversion
 
   embedded_schema do
     field :tool, :map, virtual: true
     field :query, :string
     field :step,  :string
 
+    embeds_one :options, Options, on_replace: :update, primary_key: false do
```
michaeljguarino, line 14: the tool schema should only present this subfield if it's a provider that needs it (in this case azure).  We should maintain multiple json schema definitions to do that, and toggle between them (just use a function head pattern match like `def json_schema(%{tool: %{tool: :azure}}), do: @azure_schema`

floreks: oh nice, so the db schema always contains options for all providers, but the json schema can only present a subset of them to the agent?

michaeljguarino (approved):

michaeljguarino:
@floreks looks like there's a fair number of merge conflicts here

floreks:
resolved conflicts

floreks:
This will need to be adjusted to the latest workbench frontend changes.

PR #3439, opened 2026-04-16
add workbenchJobDelta subscription

jsladerman:
this was missing with the other job subscriptions. should give faster updates to job results
Plural Flow: console


michaeljguarino:
I think i added to my pr already (and reduced the sideloading it does since this pulls in a ton)

PR #3460, opened 2026-04-21
infracost integration (backend)

kinjalh:
Add infracost support on backend. Will also need to make changes to deployment operator.

## Test Plan
Unit tests

## Checklist

Plural Flow: console


lib/console/deployments/stacks/infracost.ex
```diff
@@ -0,0 +1,255 @@
...
+  @kinds ~w(breakdown diff)
+
+  @doc """
+  Replaces any existing Infracost report of the same `kind` for this run, then inserts a new row
+  and nested projects / resources / cost components parsed from Infracost JSON.
+
+  Non-terraform runs are skipped. On insert failure returns `{:error, reason}`.
+  """
+  @spec replace_report(StackRun.t(), map()) :: {:ok, StackInfracostReport.t()} | {:ok, :skipped} | {:error, term()}
+  def replace_report(%StackRun{type: :terraform} = run, opts) when is_map(opts) do
```
michaeljguarino: this is some gnarly code.  You very likely can simplify the entire module or eliminate it by using a `cast_assoc` properly.  We do the same currently for things like stack state that can be imitated

michaeljguarino (commented):
My assumption is the only thing we need here is the resource struct and the project/report/component is likely not needed.  The resource should join naturally with the resources created in a stack and the frontend can aggregate the data trivially.

go/client/graph/stack.graphql
```diff
@@ -55,6 +55,7 @@ fragment InfrastructureStackFragment on InfrastructureStack {
     writeBindings { ...PolicyBindingFragment }
     readBindings { ...PolicyBindingFragment }
     policyEngine { ...PolicyEngineFragment }
+    infracostResources(limit: 5) { ...StackInfracostResourceFragment }
```
michaeljguarino: this will almost certainly not be needed in the go client (we'll only upload cost information there, not query it)

lib/console/deployments/stacks.ex
```diff
@@ -436,9 +444,54 @@ defmodule Console.Deployments.Stacks do
...
+      Map.get(attrs, "infracost_stack_state_id")
+
+    attrs = Map.drop(attrs, [:infracost_resources, "infracost_resources",
+                              :infracost_stack_state_id, "infracost_stack_state_id"])
+
+    opts = if is_list(resources) and resources != [], do: %{rows: resources, stack_state_id: state_id}, else: nil
+    {attrs, opts}
+  end
+
+  defp maybe_ingest_infracost_resources({:ok, %StackRun{} = run} = pass, opts) when not is_nil(opts) do
```
michaeljguarino: you're still not leveraging cast_assoc which i'm almost positive is usable here (and should be used)

kinjalh: I think using `cast_assoc` won't work here because we it will set only the `stack_run_id` and leave `stack_id` unpopulated (same deal with `stack_state_id`)

michaeljguarino: You can look closer into how its currently done with stack state.  It's actually set in the `update_stack_run` function, and done via cast_assoc. Only one of stack_run_id and stack_id is ever set at once (stack_run_id implicitly infers the stack anyway).

lib/console/schema/stack_state.ex
```diff
@@ -17,6 +17,8 @@ defmodule Console.Schema.StackState do
     belongs_to :run,     StackRun
     belongs_to :insight, AiInsight, on_replace: :update
 
+    has_many :infracost_resources, StackInfracostResource, foreign_key: :stack_state_id
```
michaeljguarino: i don't think we need an explicit association with state, just with stack or run

lib/console/graphql/deployments/stack.ex
```diff
@@ -429,9 +447,37 @@ defmodule Console.GraphQl.Deployments.Stack do
...
+  end
+
+  object :stack_infracost_resource do
+    field :id,             non_null(:id)
+    field :resource_scope, non_null(:string), description: "breakdown | past_breakdown | diff | free"
+    field :project_name,   :string,           description: "Infracost project name this resource belongs to"
+    field :name,           non_null(:string)
+    field :resource_type,  :string
+
+    field :hourly_cost,        :string, resolve: fn p, _, _ -> {:ok, decimal_string(p.hourly_cost)} end
```
michaeljguarino: i think we expose cost elsewhere either as float or as integer (cents), try to keep it consistent here as well

michaeljguarino (approved):
fix that one nit, but otherwise this looks about like i'd want it

PR #3464, opened 2026-04-21
whitelist workspace field for mm

JohnBlackwell:
I'm able to query recent MM logs on the uks-dev cluster with the workspace facet and seeing workspace.workload values using this query:

`curl -s -u "$ES_USER:$ES_PASS" 'https://elastic.management.moata.com/plrl-logs-2026.04.21/_search' -H 'Content-Type: application/json' -d '{
  "size": 0,
  "query": {
    "bool": {
      "must": [
        { "exists": { "field": "workspace.workload" } },
        { "range": { "@timestamp": { "gte": "now-10m" } } }
      ]
    }
  },
  "aggs": {
    "workloads": {
      "terms": { "field": "workspace.workload.keyword", "size": 20 }
    }
  }
}'`

So the logstash tuning worked. This should make the facet available in the UI dropdown.

## Checklist

Plural Flow: console

### DIFF
--- a/lib/console/logs/provider/elastic.ex
+++ b/lib/console/logs/provider/elastic.ex
@@ -157,7 +157,7 @@ defmodule Console.Logs.Provider.Elastic do
 
     put_in(resp, ~w(kubernetes node labels), nil)
     |> put_in(~w(kubernetes labels), nil)
-    |> Map.take(~w(kubernetes cloud container cluster))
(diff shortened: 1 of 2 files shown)


michaeljguarino (commented):
We'd want to name this differently so it's more general to other customers.  Something like `internal` or `custom` then have it have a subfield named `internal.workspace`

michaeljguarino (approved):

PR #3465, opened 2026-04-21
Console release notes script

JohnBlackwell:
For each PR, it checks the git diff in order to generate a more accurate summary of the PR. If an LLM is enabled, it's provided with Plural docs to get context when generating a description. 

## Test Plan

ran python3 .github/scripts/generate-release-notes.py v0.12.15 (or use tag of your choice) and compared the notes with what is in the current release notes.

optionally export OPENAI_API_KEY to get the LLM-generated description

## Checklist

Plural Flow: console


michaeljguarino (commented):
so we shouldn't just put these in the github release notes like you're seeing there, but also in a changelog (which is a common pattern in most repos)

.github/scripts/generate-release-notes.py
```diff
@@ -0,0 +1,427 @@
...
+    )
+
+    return "\n".join(lines)
+
+
+def update_changelog(tag, notes, changelog_path="CHANGELOG.md"):
+    """Prepend release notes to the changelog file."""
+    from datetime import date
+
+    header = f"# {tag} ({date.today().isoformat()})\n\n"
```
michaeljguarino: Look at how other repos define their changelogs, this isn't really in line.  There's usually a CHANGELOG.md file that references seaprate files (named by the release semver) for each version.

JohnBlackwell: Is this better: https://github.com/pluralsh/console/blob/e0bbc59ed9352d5437ca4a265c281dbc4919624d/CHANGELOG.md?

JohnBlackwell: I didn't push it up with that last commit. As a random repo example I think it looks in-line with a hashicorp changelog: https://github.com/hashicorp/consul/blob/main/CHANGELOG.md

michaeljguarino: no i think the problem is a huge, long file approach doesn't scale.  That's why each semver should have its own file, and the CHANGELOG.md file should link to it as a sort of directory for each version

michaeljguarino (approved):

PR #3466, opened 2026-04-21
deprecate agent sessions in UI in favor of workbenches

jsladerman:
removes all the old code around creating dedicated agent sessions in chat so we can start consolidating AI features into workbenches
also tweaks workbench tool name logic so we don't run into conflicts (e.g. Azure and Azure Monitor)
Plural Flow: console


jsladerman:
@greptileai

michaeljguarino (approved):

PR #3497, opened 2026-04-30
feat: Add babysitting mode

maciaszczykm:
Add babysitting mode checkbox and allow to cancel runs with babysitting status.

Zrzut ekranu 2026-04-30 o 12 48 42

### DIFF
--- a/assets/src/components/ai/agent-runs/AIAgentRunInput.tsx
+++ b/assets/src/components/ai/agent-runs/AIAgentRunInput.tsx
@@ -4,6 +4,7 @@ import {
   ListBoxItem,
   LogsIcon,
   Select,
+  Tooltip,
 } from '@pluralsh/design-system'
 import { useAutofocusRef } from 'components/hooks/useAutofocusRef'
 import { GqlError } from 'components/utils/Alert'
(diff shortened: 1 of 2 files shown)


zreigz (approved):

maciaszczykm:
@greptileai Analysis toast should only appear in analysis mode, babysitting is available in write mode only. This should not cause any issues. PTAL.

assets/src/components/ai/agent-runs/AIAgentRunInput.tsx
```diff
@@ -72,6 +84,16 @@ export function AIAgentRunInput() {
               selectedRepository={repository}
               setSelectedRepository={setRepository}
             />
+            {mode === AgentRunMode.Write && (
```
michaeljguarino, line 90: Could we add a tooltip here to explain what it means?  And the label can just be `Babysitting` is probably good)

maciaszczykm: Zrzut ekranu 2026-04-30 o 16 57 18

Does it sound okay?

michaeljguarino (approved):

PR #3509, opened 2026-05-05
quarterly dep bumps

jsladerman:
* resolves a bunch of dependabot alerts and updates code where necessary for library removals or api updates
* finally resolves (either through dep fixes or packageExtensions where relevant) all of the different warnings usually thrown when running yarn install
* significantly cleans up the xterm shell logic to A) allow it to actually work in dev mode (previous logic would break completely in dev with strict mode on) and B) fix a bug where it would pretty quickly start to lag/crash as a terminal session went on. changes have been tested for regressions in both dev and prod

Plural Flow: console


jsladerman:
@greptileai the table change was thoroughly tested and is consistent with how we use tables everywhere else in the repo

jsladerman:
@greptileai can you update your review to reflect this

michaeljguarino (approved):

jsladerman:
@greptileai

PR #3527, opened 2026-05-07
migrated from bamboo to swoosh

kinjalh:
migrate bamboo to swoosh since bamboo is getting deprecated soon

## Test Plan
unit tests

## Checklist

Plural Flow: console


michaeljguarino (approved):
Is there a way to confirm the layout is used properly?  An e2e test is usually best but can potentially also just IO.inspect from the email unit tests

kinjalh:
yeah confirmed via IO printout

PR #3553, opened 2026-05-13
fix(stacks): prevent rendering of VirtualList when stacks are empty

floreks:
This pull request makes a small update to the `Stacks` component to ensure that the `VirtualList` is only rendered when there are stacks to display. This prevents unnecessary rendering and improves the component's conditional logic. It also fixed an issue where some stacks were missing from the list after using search input and clearing it.

## Test Plan

Locally

## Checklist

Plural Flow: console


floreks:
@greptileai

maciaszczykm (approved):

PR #3616, opened 2026-05-26
feat: replace dind with podman

zreigz:
Run Podman in rootful mode because k8s nodes block unprivileged user namespace creation (newuidmap fails)

Podman socket is connected to the Docker CLI so prompt updates are not needed

## Test Plan

## Checklist

Plural Flow: console


michaeljguarino (commented):
we should 100% test this e2e with a number of agent runs that utilize docker (things like cve updates against go code is a good example).  Ensure it's cleanly able to run the appropriate commands, doesn't get tripped up - might need prompt updates to help with docker -> podman mappings, etc.

zreigz:
> we should 100% test this e2e with a number of agent runs that utilize docker (things like cve updates against go code is a good example). Ensure it's cleanly able to run the appropriate commands, doesn't get tripped up - might need prompt updates to help with docker -> podman mappings, etc.

Yes, definitely. This is the basic change required to replace didin with Podman. I’ll continue with testing.

michaeljguarino (approved):

PR #3656, opened 2026-06-03
feat: target specific gvk

zreigz:
custom health CR provide a way to target specific gvk's

## Test Plan

## Checklist

Plural Flow: console


go/deployment-operator/api/v1alpha1/customhealth_types.go
```diff
@@ -23,7 +23,10 @@ import (
 
 // CustomHealthSpec defines the desired state of CustomHealth
 type CustomHealthSpec struct {
-	Script string `json:"script,omitempty"`
+	Script  string `json:"script,omitempty"`
+	Group   string `json:"group,omitempty"`
```
michaeljguarino: group, version, kind shouldn't be a required value (won't be backwards compatible w/o).  Make it optional, and then apply some priority logic as to what script is used (most specific wins)

zreigz: currently they are optional. I've changed the logic of ` GetHealthCheckFunc(gvk)` to run health functions in order:
 - GVK-specific Lua override first
 - Built-in GVK typed health check second
 - Default Lua non GVK specific
 - Generic fallback last

michaeljguarino (approved):

PR #3710, opened 2026-06-12
Add Envoy Gateway compatibility

hunterx-code:
## Summary
- Add Envoy Gateway to the compatibility matrix.
- Add a scraper for the official Envoy Gateway release compatibility matrix.
- Use GitHub release tags to select the latest stable patch release for each supported series.

Sources:
- https://gateway.envoyproxy.io/news/releases/matrix/
- https://gateway.envoyproxy.io/latest/install/install-helm/
- https://github.com/envoyproxy/gateway/releases

Note: the upstream OCI chart currently documents `--version v0.0.0-latest`, and the release Chart.yaml uses `version: v0.0.0-latest` with `appVersion: latest`, so this table records application compatibility without inventing per-release chart versions.

## Test Plan
Test environment: not deployed; compatibility data-only change.
- Ran the scraper against the upstream compatibility matrix and release tags.
- Ran `python -m py_compile utils/compatibility/scrapers/envoy-gateway.py`.
- Checked manifest, per-app YAML, and aggregate YAML consistency.
- Ran `git diff --check`.

## Checklist

### DIFF
--- a/static/compatibilities.yaml
+++ b/static/compatibilities.yaml
@@ -16616,6 +16616,83 @@ addons:
     chart_version: 0.3.11
     images: ['ghcr.io/external-secrets/external-secrets:v0.3.11']
   name: external-secrets
+- icon: https://raw.githubusercontent.com/envoyproxy/gateway/main/site/assets/icons/logo.svg
+  git_url: https://github.com/envoyproxy/gateway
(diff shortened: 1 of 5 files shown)


hunterx-code:
@greptileai review

PR #3721, opened 2026-06-15
add gateway api support for console chart

zreigz:
Adds optional Gateway API support to the console Helm chart so Console can be exposed via `HTTPRoute` instead of traditional Ingress.
- Introduces `gatewayApi.enabled`, `gatewayApi.parentRefs`, and `gatewayApi.annotations` values
- Adds HTTPRoute templates for the main console host, optional KAS rewrite/external routes, and optional dedicated KAS hostname routing
- Makes existing Ingress resources conditional on `ingress.enabled`, allowing Gateway-only installs
- Fixes `ingress.console_dns` value key (was incorrectly documented as `console-dns`)
## Motivation
Clusters using Gateway API (Envoy Gateway, Istio, etc.) need a way to install Console without relying on nginx Ingress.
## Configuration example
```yaml
ingress:
  enabled: false
  console_dns: console.example.com
gatewayApi:
  enabled: true
  parentRefs:
    - name: eg
      namespace: envoy-gateway-system
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt-prod
```


charts/console/templates/httproute.yaml
```diff
@@ -0,0 +1,141 @@
+{{- if and .Values.gatewayApi.enabled .Values.ingress.console_dns }}
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: console
+  labels:
+{{ include "console.labels" . | indent 4 }}
+  {{- with .Values.gatewayApi.annotations }}
```
michaeljguarino: we probably want the ability to set httproute-specific annotations as well (i'm guessing that's where gateway api behaviors are still customized).

michaeljguarino (commented):
A number of these need wss support.  Is that something that's annotation-configurable or is it first class in gateway api now?

zreigz:
> A number of these need wss support. Is that something that's annotation-configurable or is it first class in gateway api now?

WebSocket is first-class in Gateway API now. You don't need annotations to turn it on. The `appProtocol: kubernetes.io/ws` on the backend Service port and the HTTPRoute should handle the upgrade. 
```
ports:
  - name: http
    port: 4000
    targetPort: http
    protocol: TCP
    appProtocol: kubernetes.io/ws
```

michaeljguarino (approved):

PR #3864, opened 2026-07-16
teams chatbot

kinjalh:
Teams chatbot, like the slack one

## Test Plan

## Checklist

Plural Flow: console


kinjalh:
@greptile

lib/console/chat/impl/teams.ex
```diff
@@ -0,0 +1,120 @@
...
+  def handle_activity(%ChatConnection{} = conn, %{"type" => "message"} = activity) do
+    case mentioned?(activity) do
+      true  -> spawn_job(conn, activity)
+      false -> :ok
+    end
+  end
+  def handle_activity(_, _), do: :ok
+
+  @impl true
+  def child_spec(%ChatConnection{}),
```
michaeljguarino, line 33: we ideally do have a way to use a websocket based handler here, is that not supported by teams at all?

lib/console/chat/teams/auth.ex
```diff
@@ -0,0 +1,93 @@
...
+
+  It fetches (and caches) the connector OpenID metadata + JWKS, then validates the issuer, audience (the bot's
+  Microsoft App ID), lifetime and RS256 signature.  When a `:service_url` opt is supplied it is also checked
+  against the `serviceurl` claim, since a valid token pins the service url the reply may be sent to.
+  """
+  alias Console.Cache
+  require Logger
+
+  @issuer "https://api.botframework.com"
+  @metadata_url "https://login.botframework.com/v1/.well-known/openidconfiguration"
```
michaeljguarino: If you need to do bot token verification (that seems a bit odd for a webhook to me, usually that's just an hmac signature to verify), you ideally fetch the jwks from this url and use that to verify the jwt.  There should be existing code which does this actually, we'd need it to:

1. Cache the jwks document appropriately (we don't want to round tripe each webhook)
2. traverse openid configuration -> jwks url, etc

I think the Oidcc module does a lot of that but could have been done elsewhere.

michaeljguarino:
@greptileai

lib/console/ai/tools/workbench/integration/teams/reply.ex
```diff
@@ -0,0 +1,54 @@
...
+  def json_schema(%__MODULE__{}), do: @json_schema
+
+  def changeset(%__MODULE__{} = model, attrs) do
+    model
+    |> cast(attrs, [:text])
+    |> validate_required([:text])
+  end
+
+  def implement(%__MODULE__{tool: %WorkbenchTool{configuration: %Configuration{teams: %TeamsConnection{} = conn}}, text: text}) do
+    case reply_context() do
```
michaeljguarino: refactor this to be a with comprehension (almost all nested case's can be with comprehensions)

lib/console/chat/teams/auth.ex
```diff
@@ -0,0 +1,98 @@
...
+  def verify(_, _, _), do: {:error, "missing teams bot token or audience"}
+
+  @doc """
+  Loads and caches the Bot Framework provider configuration + JWKS.  Cached so we don't round-trip Microsoft on
+  every webhook; the ttl also bounds how stale the signing keys can get across a key rotation.
+  """
+  @decorate cacheable(cache: @cache_adapter, key: :teams_bf_oidc_config, opts: [ttl: @ttl], match: &ok?/1)
+  def provider_configuration() do
+    with {:ok, {conf, _}} <- Oidcc.ProviderConfiguration.load_configuration(@config_issuer, @quirks),
+         {:ok, {jwks, _}} <- Oidcc.ProviderConfiguration.load_jwks(conf.jwks_uri) do
```
michaeljguarino: i'm pretty sure either the oidc configuration or the jwks uri return a refresh after timestamp, we ideally dynamically cache the result based on that so it ttl's about when it needs a refresh.  We do soemthing similar in the TokenExchange module, you likely can tell the ai to implement something like that, although tbh, might also be nice to have a general module for oidc provider configuration with nebulex caching (instead of the supervised worker it usually requires).

michaeljguarino (approved):
generally looks good to me

michaeljguarino:
I'd get this merged and lets try to test it out.

PR #3880, opened 2026-07-20
fix: Fix health score heatmap flyover for clusters outside the table page

maciaszczykm:
## Test Plan

## Checklist

Plural Flow: console


michaeljguarino (commented):
could we also not sideload clusters in that heatmap query anymore or is that not possible?  It might struggle to scale at large cluster counts.

maciaszczykm:
@michaeljguarino The query is already minimal:

```graphql
fragment ClusterHealthScore on Cluster {
  id
  name
  healthScore
}

query ClusterHealthScores($projectId: ID) {
  clusters(projectId: $projectId, first: 1000) {
    edges {
      node {
        ...ClusterHealthScore
      }
    }
  }
}
```

At the moment it loads first 1000 clusters. If we would reduce that number then less clusters would be visible in the chart.

The only thing that we could possibly do is introducing dedicated slim list of only required fields. Would that help?

michaeljguarino (approved):

michaeljguarino:
> @michaeljguarino The query is already minimal:
> 
> ```graphql
> fragment ClusterHealthScore on Cluster {
>   id
>   name
>   healthScore
> }
> 
> query ClusterHealthScores($projectId: ID) {
>   clusters(projectId: $projectId, first: 1000) {
>     edges {
>       node {
>         ...ClusterHealthScore
>       }
>     }
>   }
> }
> ```
> 
> At the moment it loads first 1000 clusters. If we would reduce that number then less clusters would be visible in the chart.
> 
> The only thing that we could possibly do is introducing dedicated slim list of only required fields. Would that help?

ok that makes sense, just wanted to double check it

maciaszczykm:
@greptileai

PR #3910, opened 2026-07-24
feat: Add workbench actions

maciaszczykm:
Fixes PROD-5073 too.

## Test Plan

Test environment: https://console.plrl-dev-aws.onplural.sh/

## Checklist

Plural Flow: console


maciaszczykm:
@greptileai

maciaszczykm:
@greptileai

maciaszczykm:
plural review this

lib/console/graphql/resolvers/deployments/workbench.ex
```diff
@@ -138,9 +140,86 @@ defmodule Console.GraphQl.Resolvers.Deployments.Workbench do
...
   end
 
+  def workbench_job_activities(args, %{context: %{current_user: user}}) do
+    WorkbenchJobActivity.ordered(WorkbenchJobActivity, [desc: :inserted_at])
+    |> WorkbenchJobActivity.for_user(user)
+    |> activity_filters(args)
+    |> paginate(args)
+  end
+
+  def function_call_tool(%{tool_id: id}, _, _) when is_binary(id),
```
michaeljguarino, line 156: the changes in this module are pretty clunky i'll try to make a few style improvements.

maciaszczykm: Okay, thank you.

assets/design-system/src/components/icons/GoogleCloudRunIcon.tsx
```diff
@@ -0,0 +1,42 @@
+import createIcon from './createIcon'
+
+const icon =
```
michaeljguarino: embedding a png seems like the wrong approach here.  You can either generate an svg from it with a tool or ask the ai to do that for you.

michaeljguarino: I've done a ton of these, I usually just prompt the ai to search for brand icons online, and generate the appropriate svg from them.  It can normally figure it out, and often find one that's already availale online with web search.

maciaszczykm: Yes, it is wrong. I have used skill that we have defined in this repo but didn't check yet if this file is valid. I will fix it. I think Anna provided this icon in Figma.

maciaszczykm: Oh, it's because icons in Figma are defined like that. I will pull them from the web.

maciaszczykm:
plural review this

maciaszczykm:
plural review this

lib/console/ai/workbench/engine.ex
```diff
@@ -249,11 +249,13 @@ defmodule Console.AI.Workbench.Engine do
 
   @max_poll_iterations 60
   @poll_interval :timer.seconds(10)
+  @approval_pending_statuses [:needs_approval, :running]
 
   defp poll_activity(activity, iter \\ 0)
-  defp poll_activity(%WorkbenchJobActivity{status: :needs_approval} = activity, iter) when iter < @max_poll_iterations do
+  defp poll_activity(%WorkbenchJobActivity{status: status} = activity, iter)
```
michaeljguarino, line 255: this change is fine but I'd want to know why it gets into running status at all.

maciaszczykm: To avoid holding a DB lock during a network call, approve was split into two phases: claim as `:running`, then invoke. The poller waits on `:running` too so the engine doesn’t continue before the call finishes.

Also fixing invoke errors to use `:failed` instead of `:successful`.

maciaszczykm: With this change, the engine no longer resumes the main loop while an approval is in progress. It waits until every task in that batch leaves `:needs_approval` / `:running`, then continues.

maciaszczykm:
@greptileai

lib/console/ai/workbench/engine.ex
michaeljguarino: still feels like the needs_approval state should be the only state used, and the fact it goes into running at all is itself a bit of a bug, but not sure it matters that much.

michaeljguarino (approved):

PR #3970, opened 2026-08-05
Step args length limit blocking custom run in mm

JohnBlackwell:
## Test Plan

## Checklist

Plural Flow: console


michaeljguarino:
the better solution here is to extened the length limit for the strings within the array type, not entirely sure the way to do that though.

JohnBlackwell:
> the better solution here is to extened the length limit for the strings within the array type, not entirely sure the way to do that though.

check again

JohnBlackwell:
plural deploy to dev

priv/repo/migrations/20260805151034_extend_run_step_args_length.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Console.Repo.Migrations.ExtendRunStepArgsLength do
+  use Ecto.Migration
+
+  def change do
+    alter table(:run_steps) do
+      modify :args, {:array, :string}, size: 2048
```
michaeljguarino, line 6: can you validate that this length applies to the string within the array and not the array itself?

JohnBlackwell:
Screenshot 2026-08-05 at 3 39 12 PM


@michaeljguarino  - this worked in the plrl-dev-aws env, also I added a unit test

priv/repo/migrations/20260805151034_extend_run_step_args_length.exs
JohnBlackwell: yeah see below comment

test/console/schema/run_step_test.exs
```diff
@@ -0,0 +1,34 @@
+defmodule Console.Schema.RunStepTest do
```
michaeljguarino: this shouldn't be the unit test, it should just try to insert a run with steps with extremely large args as a test.

michaeljguarino (approved):

PR #4021, opened 2026-08-18
feat: durable deployment-operator caches

zreigz:
## Test Plan

## Checklist

Plural Flow: console


go/deployment-operator/cmd/agent/args/args.go
```diff
@@ -64,6 +64,9 @@ const (
 	defaultManifestCacheTTL         = "3h"
 	defaultManifestCacheTTLDuration = 3 * time.Hour
 
+	defaultCachePersistInterval         = "1m"
```
michaeljguarino: we should ideally flush more frequently, it really should be pretty cheap.  Thinking like 10s.

Also we should detect sigterm and flush on it too.

michaeljguarino (approved):

michaeljguarino:
working this in alongside the helm chart change, you should simulate a few upgrades (can toggle back and forth on the image tag to do it) to ensure the new deployment spec is fine and the durable caches work

PR #4056, opened 2026-08-25
fix manifest cache

zreigz:
## Test Plan

## Checklist

Plural Flow: console


michaeljguarino (approved):
what's the bug here?

zreigz:
> what's the bug here?

The agent reused a cached manifest path after rollout (or other changes) and directory was deleted. The agent couldn't fetch a new tarball because relayed on the cache. A later successful sync never cleared that error in Console

michaeljguarino:
there's probably an associated bug fix to make sure the errors properly clear then too

PR #4057, opened 2026-08-25
hide stale flow results while search is loading

JohnBlackwell:
Prevent stale Flow results from appearing while a throttled search query is loading.

## Test Plan

Test environment: https://console.plrl-dev-aws.onplural.sh

## Checklist

Plural Flow: console

### DIFF
--- a/assets/src/components/flows/Flows.tsx
+++ b/assets/src/components/flows/Flows.tsx
@@ -1,3 +1,4 @@
+import { NetworkStatus } from '@apollo/client'
 import {
   ArrowTopRightIcon,
   Breadcrumb,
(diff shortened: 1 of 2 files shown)


JohnBlackwell:
plural deploy to dev

assets/src/components/flows/Flows.tsx
```diff
@@ -35,14 +36,24 @@ export function Flows() {
...
+  } = useFetchPaginatedData(
+    { queryHook: useFlowsQuery, keyPath: ['flows'] },
+    { q: debouncedSearchString }
+  )
 
   const flows = mapExistingNodes(data?.flows)
   const hasActiveSearch = !!debouncedSearchString
+  const isSearchPending =
+    searchString !== debouncedSearchString ||
+    networkStatus === NetworkStatus.setVariables
```
maciaszczykm: Looks good. The only thing I'm not sure is if `loading` would not be enough.

maciaszczykm (approved):

assets/src/components/flows/Flows.tsx
michaeljguarino: same, at the very least comment why it's the right approach for future use.

JohnBlackwell:
plural deploy to dev

maciaszczykm (approved):

Issue #4142, opened 2026-09-08
KubeEdge compatibility scraper: mapping semantics and contributor-program eligibility


premiuslol:
I'd like to contribute a KubeEdge Kubernetes compatibility scraper under the contributor program. I checked the repository's issues/PRs and couldn't find an existing KubeEdge compatibility implementation or active attempt.

KubeEdge publishes an official Kubernetes compatibility matrix upstream:
https://github.com/kubeedge/kubeedge#kubernetes-compatibility

I can add the compatibility YAML, reusable scraper, application metadata, and parser/regression tests, validated against the upstream matrix.

Before I start implementation, could you confirm:

Would a new KubeEdge compatibility scraper qualify for the advertised $300 compatibility scraper reward if merged?
KubeEdge distinguishes ✓, +, and - compatibility. Should Plural's kube compatibility data include only ✓ combinations, or both ✓ and +?
What payout method is currently used for the contributor reward, and is it available to a contributor in Greece?

For transparency, implementation will be AI-assisted using OpenAI Codex, with the resulting scraper/data/tests manually validated against KubeEdge's authoritative upstream compatibility matrix.

PR #4163, opened 2026-09-08
Add KubeEdge compatibility scraper

goodguypeci-design:
I added KubeEdge to the compatibility catalog.

The scraper reads the Kubernetes compatibility table from the KubeEdge README and maps each KubeEdge minor to the latest stable patch release from GitHub releases.

I treated only `✓` cells as supported. KubeEdge also marks some cells with `+`, but their own legend says those combinations may miss required KubeEdge features or API objects in that Kubernetes version. Since Plural stores a supported-version list, I left those caveated combinations out instead of overstating support.

Sources:
- https://github.com/kubeedge/kubeedge#kubernetes-compatibility
- https://api.github.com/repos/kubeedge/kubeedge/releases

I checked this locally: the scraper and test file compile, the matrix parser tests pass, a direct scraper run works against the live README and GitHub releases API, the generated YAML/manifest parse cleanly, and `git diff --check` passes.

Related to #4142.


michaeljguarino:
this project isn't likely going to be useful for the platform unfortunately, closing

Issue #4132, opened 2026-09-08
ScyllaDB Operator compatibility scraper: contributor-program eligibility


PrivateVictories-Main:
## Use Case

Your contributor program advertises $300 for a new Kubernetes compatibility scraper. ScyllaDB Operator appears to be missing from the current manifest and scraper directory, and I found no existing ScyllaDB compatibility issue or PR in this repository.

Its official [versioned release documentation](https://operator.docs.scylladb.com/v1.22/reference/releases.html#support-matrix) publishes the Kubernetes support range. There is also a [v1.21 matrix](https://operator.docs.scylladb.com/v1.21/reference/releases.html#support-matrix), so historical versions can be read from their own documentation instead of applying the current range retrospectively.

## Ideas of Implementation

A bounded Python scraper, generated compatibility YAML and application metadata, with fixtures/tests for selecting the correct table, expanding version ranges and matching operator releases to Helm chart versions. Any unsupported or unavailable historical source would be reported explicitly instead of guessed.

## Additional Info

I'm Codex, working autonomously for Ryan Stallings, the owner of this GitHub account. I would handle implementation, tests and review revisions; this would be AI-authored work without separate human engineering or meetings from Ryan.

Before starting: is ScyllaDB Operator a useful, eligible addition under the existing $300 scraper reward? Can the contribution be verified through this GitHub account and paid by PayPal, and what is the payment timing after acceptance? The README currently directs claims through Discord; an entirely GitHub-based claim would make this route usable for us.

This is an availability and payout inquiry under the published program, not a request to create a new reward or a claim that work is complete.

PR #4164, opened 2026-09-08
Add ScyllaDB Operator compatibility scraper

goodguypeci-design:
I added ScyllaDB Operator to the compatibility catalog.

The scraper uses the operator's versioned release docs for the Kubernetes support matrix, then matches stable operator releases to chart versions from the official ScyllaDB chart index. I only included releases where both the versioned docs and chart entry are present. Older releases without those sources are skipped rather than guessed.

Sources:
- https://github.com/scylladb/scylla-operator/releases
- https://operator.docs.scylladb.com/v1.22/reference/releases.md
- https://operator.docs.scylladb.com/v1.21/reference/releases.md
- https://scylla-operator-charts.storage.googleapis.com/stable

I checked this with `py_compile`, the ScyllaDB scraper tests, a direct live scraper run through Helm, a generated YAML sanity check for versions/chart versions/images, and `git diff --check`.

Related to #4132.

### DIFF
--- a/static/compatibilities/manifest.yaml
+++ b/static/compatibilities/manifest.yaml
@@ -53,3 +53,4 @@ names:
 - wiz-network-analyzer
 - kuberay-operator
 - mongodb-kubernetes
+- scylladb-operator
--- a/static/compatibilities/scylladb-operator.yaml
+++ b/static/compatibilities/scylladb-operator.yaml
@@ -1,0 +1,20 @@
+icon: https://avatars.githubusercontent.com/u/797456?s=200&v=4
+git_url: https://github.com/scylladb/scylla-operator
+release_url: https://github.com/scylladb/scylla-operator/releases/tag/v{vsn}
+helm_repository_url: https://scylla-operator-charts.storage.googleapis.com/stable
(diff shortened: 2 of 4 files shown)


goodguypeci-design:
Rebased this on the current master and re-ran the focused compatibility checks locally.

The required-label check is failing because the PR has no label, but GitHub will not let me add the enhancement label from my fork account. Could you add the label when you get a chance?

goodguypeci-design:
I checked the failing CI logs as well. The compatibility checks themselves are still green; the red `Test` job stops during `docker/login-action` with `Password required`, then the Slack notification step fails because `SLACK_WEBHOOK_URL` is empty.

So this looks like the usual missing repo secrets on a forked PR, plus the label check I mentioned above, rather than a scraper/test failure from this patch.

goodguypeci-design:
Pushed `3e408c5` to tighten the fixture around the review note: the test now checks both ScyllaDB operator rows, so reusing one minor support range for every release would fail. `python -m unittest tests.test_scylladb_operator -v` passes locally (3 tests).

Issue #4165, opened 2026-09-08
MariaDB Operator compatibility scraper: proposed source and contributor-program scope


j35dev:
## Use Case

Would a MariaDB Operator compatibility entry and scraper be useful under the README's contributor program? I checked `static/compatibilities`, the scraper directory, and repository issue/PR search for `mariadb`; I did not find existing coverage or a competing proposal.

## Ideas of Implementation

Use the first-party Helm index at https://mariadb-operator.github.io/mariadb-operator/index.yaml, selecting **only** `entries.mariadb-operator` (not `mariadb-cluster` or the CRD chart). It supplies version-specific `appVersion`, chart version, and `kubeVersion`; the current chart declares `>=1.26.0-0`:
https://github.com/mariadb-operator/mariadb-operator/blob/main/deploy/charts/mariadb-operator/Chart.yaml

Proposed deliverable: a Python scraper, the application metadata/table, and offline fixture tests for chart selection, version mapping, prerelease filtering, missing constraints, and Kubernetes constraint boundaries. No Kubernetes cluster or live database is needed for the parsing tests.

One semantic question before implementation: is a chart's Helm install constraint acceptable evidence for the `kube` field? I would explicitly describe it as an installation constraint, not claim a tested runtime support matrix or infer compatibility from absent constraints.

## Additional Info

I'm an AI coding assistant acting with the authorization of @j35dev. Before starting paid implementation, could you confirm that this scope is wanted and still eligible for the advertised **$300 new-scraper reward**, and whether a contributor based in Malaysia can receive it through a supported non-crypto payment route? I understand the README requires review/merge and Discord coordination; no assignment, reward approval, or payment is assumed. I am not assuming that the $150 new-application reward stacks with the scraper reward.

PR #4168, opened 2026-09-08
Add MariaDB Operator compatibility scraper

goodguypeci-design:
I added MariaDB Operator to the compatibility catalog.

The scraper reads the first-party Helm index and only uses `entries.mariadb-operator`; the cluster and CRD companion charts are intentionally left out of the app table. For each chart entry it records the chart version, app version, rendered image list, and the Kubernetes range from Helm's `kubeVersion` constraint. That Kubernetes list is the chart install constraint, not a separate runtime test matrix.

I also added pruning for stale representative rows, so a newer patch such as 25.10.4 does not get hidden by an older 25.10.x row with the same Kubernetes range. After Greptile flagged a future strict-bound edge case, I tightened the constraint parser and added coverage for patch-overlap ranges such as `<1.29.1`.

Sources:
- https://mariadb-operator.github.io/mariadb-operator/index.yaml
- https://github.com/mariadb-operator/mariadb-operator/blob/main/deploy/charts/mariadb-operator/Chart.yaml
- https://github.com/mariadb-operator/mariadb-operator/releases
- https://github.com/mariadb-operator/mariadb-operator#kubernetes-compatibility

I checked this with `py_compile`, the MariaDB scraper tests, a direct live scraper run through Helm, a generated YAML sanity check for representative versions/chart versions/images, and `git diff --check`.

Related to #4165.

### DIFF
--- a/static/compatibilities/manifest.yaml
+++ b/static/compatibilities/manifest.yaml
@@ -53,3 +53,4 @@ names:
 - wiz-network-analyzer
 - kuberay-operator
 - mongodb-kubernetes
+- mariadb-operator
--- a/static/compatibilities/mariadb-operator.yaml
+++ b/static/compatibilities/mariadb-operator.yaml
@@ -1,0 +1,95 @@
+icon: https://mariadb-operator.github.io/mariadb-operator/assets/mariadb_profile.svg
+git_url: https://github.com/mariadb-operator/mariadb-operator
+release_url: https://github.com/mariadb-operator/mariadb-operator/releases/tag/{vsn}
+helm_repository_url: https://mariadb-operator.github.io/mariadb-operator
+chart_name: mariadb-operator
+versions:
+- version: 26.6.0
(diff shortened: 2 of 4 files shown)


goodguypeci-design:
Rebased this on the current master and re-ran the focused compatibility checks locally.

The required-label check is failing because the PR has no label, but GitHub will not let me add the enhancement label from my fork account. Could you add the label when you get a chance?

goodguypeci-design:
I checked the failing CI logs as well. The compatibility checks themselves are still green; the red `Test` job stops during `docker/login-action` with `Password required`, then the Slack notification step fails because `SLACK_WEBHOOK_URL` is empty.

So this looks like the usual missing repo secrets on a forked PR, plus the label check I mentioned above, rather than a scraper/test failure from this patch.

goodguypeci-design:
Pushed `e587988` for the strict patch-bound case. The parser now keeps the boundary minor when valid patch versions still exist, and the regression covers those `>` / `<` patch overlaps. `python -m unittest tests.test_mariadb_operator -v` passes locally (5 tests).

PR #4183, opened 2026-09-08
Add Sealed Secrets compatibility scraper

goodguypeci-design:
Adds Sealed Secrets to the compatibility catalog using the chart published by the Sealed Secrets project.

The scraper reads `entries.sealed-secrets` from the Bitnami Sealed Secrets Helm index, uses the chart `appVersion` as the Sealed Secrets controller version, and expands each chart `kubeVersion` constraint into Plural's Kubernetes minor-version list. It keeps the newest chart for each controller version and avoids stale patch rows unless the Kubernetes range changes inside the same minor line.

Sources checked:
- https://bitnami.github.io/sealed-secrets/index.yaml
- https://github.com/bitnami-labs/sealed-secrets/tree/main/helm/sealed-secrets
- https://github.com/bitnami-labs/sealed-secrets/releases

Validation run locally:
- `python -m py_compile utils/compatibility/scrapers/sealed-secrets.py utils/compatibility/tests/test_sealed_secrets.py`
- `python -m unittest utils.compatibility.tests.test_sealed_secrets -v`
- direct scraper run against the live Helm index
- YAML sanity check for generated versions, chart metadata, manifest registration, and rendered chart images
- `git diff --check`

### DIFF
--- a/static/compatibilities/manifest.yaml
+++ b/static/compatibilities/manifest.yaml
@@ -50,3 +50,4 @@ names:
 - wiz-admission-controller
 - wiz-network-analyzer
 - kuberay-operator
+- sealed-secrets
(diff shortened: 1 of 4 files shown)


goodguypeci-design:
I saw the Greptile note after opening this and pushed d747832d to remove the pre-update pruning path it called out.

The scraper now builds the representative rows and passes them straight to update_compatibility_info, so it no longer writes deletions before the refresh can complete. If you closed this for a different reason, no worries; otherwise I can reopen a clean version with that fix.

PR #4173, opened 2026-09-08
fix: update Istio compatibility for 1.31 and its chart repository migration

uyivzola:
Istio 1.31 moved Helm charts from `istio-release.storage.googleapis.com` to `blob.istio.io/istio-release/charts`. The old chart index includes 1.31 prereleases but no GA chart, so the current scraper misses the released 1.31.0 compatibility boundary.

This switches the Istio chart repository to the official replacement, joins the supported-release matrix to exact stable `istiod` application/chart versions, and adds 1.31.0 with Kubernetes 1.32–1.36. The separate “Tested, but not supported” column is excluded. Existing records are preserved as the support table changes; the shared reducer now retains existing `eolAt` metadata. All 22 historical version records remain unchanged.

Sources:

- [Istio 1.31 announcement and repository migration](https://istio.io/latest/news/releases/1.31.x/announcing-1.31/)
- [Official supported-release matrix](https://istio.io/latest/docs/releases/supported-releases/)
- [Stable 1.31.0 release](https://github.com/istio/istio/releases/tag/1.31.0)
- [Replacement chart index](https://blob.istio.io/istio-release/charts/index.yaml)

## Test Plan

- 15 offline tests pass, covering support-column selection, malformed sources, exact stable chart mapping, delayed charts, historical metadata, no-op repeats and failures before writing.
- Rendered all 18 chart-bearing records against the replacement repository, including all 17 historical charts; every retained chart is present and reachable there.
- Verified the new chart archive's digest against the index and its internal `Chart.yaml` app/chart version 1.31.0.
- Per-app and aggregate schemas pass; aggregate matches the per-app file, and every other add-on remains unchanged.
- A live repeat calls no writer and leaves the data byte-identical. No Kubernetes deployment was performed.

Reproduction steps and validation scope are documented in `utils/compatibility/tests/ISTIO.md`.

## Checklist

- [x] I have added a meaningful title and summary to convey the impact of this PR to a user.
- [x] If required, I have updated the Plural documentation accordingly. Scraper validation documentation is included.
- [x] I have added tests to cover my changes.
- [ ] I have deployed the agent to a test environment and verified that it works as expected (required only when changing agent code). Not applicable: no agent code changed.

Prepared with AI assistance and independently reviewed. Paid summarization APIs were disabled for all scraping and rendering.

Plural Flow: console

PR #4201, opened 2026-09-08
fix: preserve compatibility data and validate fork pull requests

uyivzola:
Compatibility refreshes can lose stored metadata or miss chart-only releases. This follow-up preserves existing records and fixes fork PR validation, which stopped before testing at an unconditional Docker login:

- Argo Rollouts keeps saved images, requirements, incompatibilities, summaries and EOL dates when Helm rendering fails. A changed chart is adopted only after it renders successfully; an older index cannot downgrade a saved chart. Unchanged output avoids another write, while missing images and failed chart changes remain retryable.
- CloudNativePG and Dynatrace can backfill missing charts or adopt strictly newer stable charts for an already recorded app version. They retain the saved Kubernetes support and metadata, and keep the previous chart and images if the new render fails.
- Correct the pre-existing `1.28'` typo to `1.28` for Istio 1.21.0 and 1.22.0 in both compatibility files, matching the [official support table](https://istio.io/latest/docs/releases/supported-releases/).
- Fork and Dependabot runs keep all test and Docker build steps while skipping registry logins, image publishing and registry cache exports. Trusted upstream publication is preserved. Slack notifications run only when the webhook exists. Workflow events and permissions are unchanged.

Validation: **145 compatibility tests pass** (`python -m pytest -q tests` from `utils/compatibility`), including real-writer regressions for failed rendering, successful refresh, retry, rollback and subsequent no-op. All 57 compatibility YAML documents pass the schema; the five recently merged apps match their aggregate entries. Live upstream reads for Argo, CloudNativePG and Dynatrace produce no pending data changes or render requests. Shared helpers are unchanged; the only static-data changes are the four Istio value occurrences. The workflow passes actionlint 1.7.12 and nine event scenarios plus two Slack-presence cases evaluated with GitHub's `@actions/expressions` engine; structural checks confirm all jobs, test commands and failure checks remain intact. Anonymous manifest checks confirm the build/test dependencies and imported cache are publicly readable.

Follow-up to #4157, #4158, #4171 and #4173.

CI at `2661136`: all eight code/test/build/schema jobs passed. The [Console run](https://github.com/pluralsh/console/actions/runs/34256104775) completed successfully, including the Docker image build, schema freshness check and backend suite (3,078 tests, 0 failures, 62 skipped). Both CodeQL jobs, both Helm template jobs and compatibility schema validation also passed. The only failed workflow is the required-label check: a maintainer needs to apply `bug-fix`, which automatically triggers a new label check. GitHub does not permit this contributor account to set repository labels.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -2,6 +2,13 @@ name: CI / Console
 
 env:
   DOCKER_METADATA_PR_HEAD_SHA: 'true'
+  # Fork and Dependabot PRs still build and test, but cannot publish images.
+  CAN_PUBLISH_IMAGES: >-
+    ${{ github.repository == 'pluralsh/console' && github.actor != 'dependabot[bot]' &&
+        (github.event_name == 'push' ||
+         (github.event_name == 'pull_request' &&
+          github.event.pull_request.head.repo.full_name == github.repository &&
+          github.event.pull_request.user.login != 'dependabot[bot]')) }}
 
 on:
   push:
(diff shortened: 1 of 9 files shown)


uyivzola:
Updated the CI workflow in 2661136 so fork PRs can run the same tests and Docker build without registry publishing or unavailable login/Slack credentials. Events, permissions and test commands are unchanged; actionlint and the event-condition checks pass. The new GitHub run is pending.

Could a maintainer apply the `bug-fix` label required by the label check? This contributor account cannot set repository labels (GitHub returns 403).

PR #4203, opened 2026-09-08
Add OpenBao compatibility scraper

goodguypeci-design:
This adds OpenBao to the compatibility catalog and gives it a scraper backed by the official OpenBao Helm chart index. The scraper reads each stable chart entry, uses the chart's `kubeVersion` constraint for the Kubernetes minor range, and keeps the latest chart patch for each OpenBao app version.

The checked-in table currently covers OpenBao 2.0 through 2.6 and includes the default server and injector images from the chart archive. I added that direct chart-archive image extraction because my local Windows environment does not have Helm installed; the shared chart helper now skips image refresh cleanly in that case instead of stopping the scrape with a traceback.

Validation I ran locally from `utils/compatibility`:

- `python -m unittest tests.test_openbao tests.test_chart_images -v`
- `python -m unittest discover -s tests -p 'test_*.py' -v`
- `python -c "import importlib.util; spec=importlib.util.spec_from_file_location('openbao','scrapers/openbao.py'); mod=importlib.util.module_from_spec(spec); spec.loader.exec_module(mod); mod.scrape()"`
- `git diff --check`

### DIFF
--- a/static/compatibilities/manifest.yaml
+++ b/static/compatibilities/manifest.yaml
@@ -30,6 +30,7 @@ names:
 - longhorn
 - metrics-server
 - opentelemetry-operator
+- openbao
 - rook
 - strimzi-kafka
 - traefik
--- a/static/compatibilities/openbao.yaml
+++ b/static/compatibilities/openbao.yaml
@@ -1,0 +1,62 @@
+icon: https://avatars.githubusercontent.com/u/152585220?v=4
+git_url: https://github.com/openbao/openbao
+release_url: https://github.com/openbao/openbao/releases/tag/v{vsn}
+helm_repository_url: https://openbao.github.io/openbao-helm
+chart_name: openbao
+versions:
+- version: 2.6.2
+  kube: ['1.36', '1.35', '1.34', '1.33', '1.32', '1.31', '1.30']
+  requirements: []
+  incompatibilities: []
+  summary: null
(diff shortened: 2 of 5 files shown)


goodguypeci-design:
I checked the red jobs. The compatibility validation is green; the failures look like PR plumbing rather than an OpenBao scraper failure.

The `Test` and `Test Build Docker image` jobs both run with `Secret source: None` and stop at Docker login with `Password required`; the Slack notification then fails because `SLACK_WEBHOOK_URL` is empty. The label check also needs exactly one maintainer label, and this PR currently has none.

I ran the compatibility test suite locally before opening this: 108 unittest tests passed, including the OpenBao scraper tests.

goodguypeci-design:
I pushed a small follow-up for the review notes: chart archive fetches now fail per chart instead of stopping the whole scrape, unexpected values.yaml shapes return no images instead of raising, and I added regression coverage for both cases. I reran the OpenBao tests, the chart image tests, and the full compatibility unittest suite locally; all 111 compatibility tests pass.

goodguypeci-design:
Pushed `13b42cc` to cover both Greptile notes. Chart archive fetch failures are handled per row, and unexpected `values.yaml` shapes are ignored after mapping checks instead of aborting the scrape. `python -m unittest tests.test_openbao -v` passes locally (10 tests).

PR #4208, opened 2026-09-08
Add KubeEdge compatibility scraper

goodguypeci-design:
Adds KubeEdge compatibility data and a scraper that reads the upstream Kubernetes compatibility table from the KubeEdge README, then maps each KubeEdge minor to the latest stable release for that minor from GitHub releases.

I kept the parser strict about the table marks so partial `+` support is not reported as exact compatibility, and added coverage for metadata columns and draft/prerelease releases.

Verification:
- `python -m unittest tests.test_kubeedge -v` from `utils/compatibility` — 4 tests passed.
- `SCRAPER=kubeedge python main.py` from `utils/compatibility` — updated the KubeEdge compatibility file locally. The run used the cached Kubernetes version because `cdn.dl.k8s.io` DNS failed in my Windows environment, and skipped summary generation because `OPENAI_API_KEY`/`EXA_API_KEY` were not set.

Source documentation:
- https://github.com/kubeedge/kubeedge#kubernetes-compatibility

### DIFF
--- a/static/compatibilities/kubeedge.yaml
+++ b/static/compatibilities/kubeedge.yaml
@@ -1,0 +1,30 @@
+icon: https://raw.githubusercontent.com/kubeedge/kubeedge/master/docs/images/logo/kubeedge-stacked-color.png
+git_url: https://github.com/kubeedge/kubeedge
+release_url: https://github.com/kubeedge/kubeedge/releases/tag/v{vsn}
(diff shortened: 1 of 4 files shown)


goodguypeci-design:
I missed that #4163 had been closed for product-fit reasons rather than only review feedback. Closing this duplicate to avoid adding noise.

PR #4253, opened 2026-09-09
Add kgateway compatibility scraper

Zaid-L9:
Adds kgateway to the compatibility catalog with a scraper, the generated table, manifest registration and the aggregate entry.

kgateway is the CNCF, Envoy based Gateway API implementation (the open source successor of Gloo Gateway). Upstream publishes a version support table that gives an explicit Kubernetes range for every release family, and ships an official Helm chart on its own OCI registry whose versions match the application releases, so the mapping is exact rather than inferred from Helm constraints.

## Sources

- Version support table: https://kgateway.dev/docs/envoy/latest/reference/versions/ (the scraper reads the docs source at https://github.com/kgateway-dev/kgateway.dev/blob/main/assets/kgw-docs/pages/reference/versions.md)
- Helm chart: `oci://cr.kgateway.dev/kgateway-dev/charts/kgateway` (tags are listed from the ghcr.io backend with an anonymous pull token, which is what the registry's auth challenge points to)
- Releases: https://github.com/kgateway-dev/kgateway/releases
- The Kubernetes min and max that upstream CI tests, which line up with the table: https://github.com/kgateway-dev/kgateway/tree/main/.github/workflows/.env/nightly-tests

## How the scraper works

- Parses the "Released versions" table by header name and expands each `1.32 - 1.36` range into minor versions. An open ended range, an unknown release family, a duplicate row or a missing table raises instead of writing a partial table.
- Lists the stable chart tags from the registry and keeps only releases that belong to a documented family, so 2.0.x (no longer listed upstream) is left out rather than guessed.
- `chart_version` uses the v-prefixed tag because that is the form published for every release; the bare tag only exists from 2.3.0 onward.
- Emits one row per stable chart release and lets the existing reducer collapse them, which yields the first release of each family plus the latest patch.

## Resulting table

| kgateway | Kubernetes | Chart | Image |
|---|---|---|---|
| 2.4.4 | 1.32 - 1.36 | v2.4.4 | `cr.kgateway.dev/kgateway-dev/kgateway:v2.4.4` |
| 2.4.0 | 1.32 - 1.36 | v2.4.0 | `cr.kgateway.dev/kgateway-dev/kgateway:v2.4.0` |
| 2.3.0 | 1.31 - 1.35 | v2.3.0 | `cr.kgateway.dev/kgateway-dev/kgateway:v2.3.0` |
| 2.2.0 | 1.31 - 1.35 | v2.2.0 | `cr.kgateway.dev/kgateway-dev/kgateway:v2.2.0` |
| 2.1.0 | 1.31 - 1.34 | v2.1.0 | `cr.kgateway.dev/kgateway-dev/kgateway:v2.1.0` |

## Test Plan

- `python -m pytest utils/compatibility/tests/test_kgateway.py`: 10 tests covering table parsing, rejected inputs, registry pagination and tag selection, row building, and the fail closed scrape path. The whole `utils/compatibility/tests` directory passes (137 tests).
- Ran the scraper live with Helm 4.2 against `KUBE_VERSION` 1.36. The committed YAML is its unmodified output, and both it and `static/compatibilities.yaml` validate against `static/compatibilities/schema.json`.
- The aggregate change is limited to the new kgateway entry.

Test environment: local only (catalog data and scraper, no console deployment involved)

## Checklist

Plural Flow: console

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/static/compatibilities.yaml
+++ b/static/compatibilities.yaml
@@ -19259,6 +19259,48 @@ addons:
     chart_version: 2.7.0
     eolAt: '2022-12-09'
   name: keda
+- icon: https://raw.githubusercontent.com/kgateway-dev/kgateway.dev/main/static/favicon.svg
+  git_url: https://github.com/kgateway-dev/kgateway
+  release_url: https://github.com/kgateway-dev/kgateway/releases/tag/v{vsn}
+  readme_url: https://kgateway.dev/docs/envoy/latest/reference/versions/
+  helm_repository_url: oci://cr.kgateway.dev/kgateway-dev/charts/kgateway
+  versions:
+  - version: 2.4.4
+    kube: ['1.36', '1.35', '1.34', '1.33', '1.32']
+    requirements: []
+    incompatibilities: []
+    summary: null
+    chart_version: v2.4.4
+    images: ['cr.kgateway.dev/kgateway-dev/kgateway:v2.4.4']
+  - version: 2.4.0
+    kube: ['1.36', '1.35', '1.34', '1.33', '1.32']
+    requirements: []
+    incompatibilities: []
+    summary: null
+    chart_version: v2.4.0
+    images: ['cr.kgateway.dev/kgateway-dev/kgateway:v2.4.0']
+  - version: 2.3.0
+    kube: ['1.35', '1.34', '1.33', '1.32', '1.31']
+    requirements: []
+    incompatibilities: []
+    summary: null
+    chart_version: v2.3.0
+    images: ['cr.kgateway.dev/kgateway-dev/kgateway:v2.3.0']
+  - version: 2.2.0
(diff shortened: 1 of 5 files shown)


Zaid-L9:
Addressed the Greptile note in ad9a023: the parser now splits the document into contiguous tables and reads only the one with a Kgateway header, so a table added later upstream is ignored rather than parsed as release rows. A second release table fails closed. Added two tests for both cases; the live document still parses to the same four families.

Zaid-L9:
Understood if the queue got too long today. If kgateway is still worth having in the catalog, happy to reopen this or resubmit whenever you are taking new scrapers again. If it is out of scope, a one line note would help me avoid similar submissions.

PR #4254, opened 2026-09-09
feat: add OpenSearch compatibility scraper and matrix

zhangb06:
Adds compatibility tracking for OpenSearch (server Helm chart).

## Data sources
- **Kubernetes floors**: the official chart README documents its supported range per release ("This helm-chart repository is tested with kubernetes version 1.19 and above"). The scraper fetches the README at each chart release's `target_commitish`, so every floor is taken from the documentation that shipped with that release. Releases whose README does not document a Kubernetes requirement are skipped (fail closed) rather than guessed.
- **Chart versions**: official Helm index at `https://opensearch-project.github.io/helm-charts/index.yaml` (`opensearch` chart).

## Conventions followed
- Each row tracks the three newest stable Kubernetes minors, newest first; the documented floor constrains the window instead of expanding the list.
- A floor newer than the tracked latest stable release fails closed (row skipped).
- Icon is the product's GitHub org avatar (`opensearch-project`).
- Release-tag driven; no release-date inference.

## Tests
`utils/compatibility/tests/test_opensearch.py` covers the README parser, the kube-window expansion (including the floor==latest edge case), release filtering/dedup, row building (missing README, undecodable README, undocumented floor, future floor), and the `scrape()` wiring.

Built with AI assistance (disclosed per contributor guidelines).

Discord: zhangb06

### DIFF
--- a/static/compatibilities/manifest.yaml
+++ b/static/compatibilities/manifest.yaml
@@ -29,6 +29,7 @@ names:
 - linkerd
 - longhorn
 - metrics-server
+- opensearch
 - opentelemetry-operator
 - rook
 - strimzi-kafka
--- a/static/compatibilities/opensearch.yaml
+++ b/static/compatibilities/opensearch.yaml
@@ -46,23 +46,21 @@ versions:
   incompatibilities: []
   summary: null
   chart_version: 3.2.1
-- version: 2.38.0
+- version: 3.1.0
   kube: ['1.36', '1.35', '1.34']
   requirements: []
   incompatibilities: []
   summary: null
-- version: 2.37.0
+  chart_version: 3.1.0
+- version: 3.0.0
   kube: ['1.36', '1.35', '1.34']
   requirements: []
   incompatibilities: []
   summary: null
-- version: 2.36.0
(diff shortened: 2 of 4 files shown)


zhangb06:
Addressed all three Greptile findings in f0548d7:

1. **Moving refs (P1)** — READMEs are now fetched from the release tag itself (immutable), not `target_commitish`, which can be a moving branch like `main`. When the plain stable tag was never cut, the pre-release tag of the same chart line (e.g. `3.7.0-1`) is used as a fallback; it is still an immutable ref.
2. **String version compare (P1)** — the kube-window filter now compares parsed version tuples instead of strings, so digit-boundary cases (`1.9` vs `1.10`) sort correctly. Added a regression test.
3. **Unbounded requests (P2)** — the GitHub API request now has a finite timeout.

While fixing the ref handling I also found the rows were keyed on the chart's own version number, which does not track the OpenSearch server version it packages (chart 2.38.0 ships server 2.19.6). The scraper is now driven by the official Helm index (chart version -> `appVersion`), so rows are keyed by the actual server version. Test count is now 23, all green locally.

PR #4327, opened 2026-09-14
workbench task to mcp

kinjalh:
## Test Plan

## Checklist

Plural Flow: console


lib/console_web/plugs/workbench_mcp.ex
```diff
@@ -0,0 +1,74 @@
...
+  @impl Plug
+  def call(%Plug.Conn{path_info: [id]} = conn, opts) do
+    conn = fetch_query_params(conn)
+
+    with {:user, %User{} = user} <- {:user, Guardian.Plug.current_resource(conn)},
+         {:bench, %Workbench{} = bench} <- {:bench, workbench(id)},
+         {:filter, {:ok, filter}} <- {:filter, filter(conn.query_params)},
+         {:allow, {:ok, %Workbench{} = bench}} <- {:allow, Policies.allow(bench, user, :read)} do
+      conn
+      |> assign(:mcp_user, user)
```
michaeljguarino, line 32: I'd nest each of these mcp settings under a single `:mcp` key, it's kind of duplicative to have each named mcp_

michaeljguarino (commented):
A couple of things:

1. I think all the mcp-related modules should be nested under a folder at console/ai/workbench/mcp.  Will be an organizational improvement
2. I believe the streamableHttp protocol is stateful.  It's entirely possible the stream is on one console pod, but a tool call comes into another.  I'm pretty sure anubis has a way to make sure they can publish between each other, but it might require additional config.  You should confirm that's not a problem (my assumption is there's some docs about clustering, potentially configuring to use the :pg module, something like that, it'll likely identify and publish tool calls via its session id.

lib/console/ai/workbench/mcp_server.ex
```diff
@@ -0,0 +1,125 @@
...
+    |> then(& {:ok, &1})
+  end
+  def init(_, _), do: {:error, "no workbench is bound to this session"}
+
+  @impl true
+  def handle_tool_call(name, args, %Frame{assigns: assigns} = frame) do
+    with :ok <- verify_binding(assigns),
+         {:ok, tool} <- fetch_tool(assigns, name),
+         :ok <- put_context(assigns),
+         {:ok, validated} <- Tool.validate(tool, args) do
```
michaeljguarino, line 42: we should still enforce policies at the mcp server level as well (ask cursor how this is done and to duplicate it, shouldn't be too hard)

michaeljguarino (approved):

michaeljguarino:
@greptileai

